Issue #1, opened 2019-02-05
[Code size] Emit __mspabi_func_epilog instead of a sequence of 'pop' instructions


chbessonova:
MSP430 EABI (see p. 3.8) provides `__mspabi_func_epilog_X` functions set to replace a sequence of `pop` instructions at the end of a function in order to reduce code size.
GCC uses the approach, `__mspabi_func_epilog_X` is implemented in libgcc. 

For example,
```
30 40 00 00 	br	#0x0000		;
        R_MSP430_16	__mspabi_func_epilog_7
```
instead of
```
34 41       	pop	r4		;
35 41       	pop	r5		;
36 41       	pop	r6		;
37 41       	pop	r7		;
38 41       	pop	r8		;
39 41       	pop	r9		;
3a 41       	pop	r10		;
```

atrosinenko:
@cr1901 In case someone else could be interested in reviewing this msp430-specific patch: https://reviews.llvm.org/D84397

cr1901:
@atrosinenko My attempt at testing is blocked pending cargo fixes. I will get to it when possible.

cr1901:
@atrosinenko 

## Before Patching
Rust commit [09f4c9f5082f78](https://github.com/rust-lang/rust/commit/09f4c9f5082f78b0adcee83d3ab4337e000cd28e), LLVM commit [86b120e6f30](https://github.com/rust-lang/llvm-project/tree/86b120e6f302d39cd6973b6391fb299d7bc22122), which _corresponds to LLVM 10_.

```
msp430-elf-size target/msp430-none-elf/release/at2xt
   text    data     bss     dec     hex filename
   1998       4      46    2048     800 target/msp430-none-elf/release/at2xt
```

## Patching
```
patch -Np1 < D84397.diff

patching file llvm/lib/Target/MSP430/CMakeLists.txt
patching file llvm/lib/Target/MSP430/MSP430.h
Hunk #1 succeeded at 43 (offset 1 line).
patching file llvm/lib/Target/MSP430/MSP430CommonEpilogueOptimizer.cpp
patching file llvm/lib/Target/MSP430/MSP430InstrInfo.cpp
Hunk #2 succeeded at 194 (offset 12 lines).
Hunk #3 succeeded at 212 (offset 12 lines).
patching file llvm/lib/Target/MSP430/MSP430TargetMachine.cpp
patching file llvm/test/CodeGen/MSP430/asm-clobbers.ll
[...]

PR #8, opened 2020-04-14
[WIP] Emit tail calls to __mspabi_func_epilog_X

atrosinenko:
Here is a draft PR for internal discussion of a fix for #1. For now, it is expected to be squashed/rebased periodically, when appropriate for discussion.

To Do
- [x] create a stub patch that just compiles and tests itself somehow
- [ ] ensure the resulting code can be linked correctly (all symbols can be resolved, etc.)
- [ ] find out why `R4` seems to not being spilled just like `R5` - `R10` (or why this is correct). For now, this looks contradicting MSP EABI
- [ ] get rid of the dangling `ret` at the very end of a function
- [ ] rewrite string handling idiomatically (symbol names: use literal constant strings, dynamic formatting on stack / on heap, etc.; speed / allocation / maintainability trade-off)
- [ ] ultimately improve it to upstreamable quality

### DIFF
--- a/llvm/lib/Target/MSP430/MSP430FrameLowering.cpp
+++ b/llvm/lib/Target/MSP430/MSP430FrameLowering.cpp
@@ -200,6 +200,31 @@ bool MSP430FrameLowering::spillCalleeSavedRegisters(
   return true;
 }
 
+const char *MSP430FrameLowering::tailCallNameForRestoringCalleSavedRegisters(MutableArrayRef<CalleeSavedInfo> CSI) const {
+  static unsigned Registers[] = {
+    /*MSP430::R4, ??? */ MSP430::R5, MSP430::R6, MSP430::R7,
+    MSP430::R8, MSP430::R9, MSP430::R10,
+  };
+  static const char *RestorerNames[] = {
+    nullptr,
+    "__mspabi_func_epilog_1",
+    "__mspabi_func_epilog_2",
+    "__mspabi_func_epilog_3",
+    "__mspabi_func_epilog_4",
+    "__mspabi_func_epilog_5",
+    "__mspabi_func_epilog_6",
+    "__mspabi_func_epilog_7",
+  };
+  const unsigned TotalCount = sizeof(Registers) / sizeof(Registers[0]);
+  const unsigned PoppedCount = CSI.size();
+  for (unsigned i = 0; i < PoppedCount; ++i) {
+    if (CSI[i].getReg() != Registers[TotalCount - PoppedCount + i]) {
+      return nullptr;
+    }
+  }
+  return RestorerNames[PoppedCount];
+}
+
 bool MSP430FrameLowering::restoreCalleeSavedRegisters(
     MachineBasicBlock &MBB, MachineBasicBlock::iterator MI,
     MutableArrayRef<CalleeSavedInfo> CSI, const TargetRegisterInfo *TRI) const {
@@ -211,6 +236,12 @@ bool MSP430FrameLowering::restoreCalleeSavedRegisters(
 
   MachineFunction &MF = *MBB.getParent();
   const TargetInstrInfo &TII = *MF.getSubtarget().getInstrInfo();
+
+  const char *RestorerName = tailCallNameForRestoringCalleSavedRegisters(CSI);
+  if (RestorerName != nullptr) {
+    BuildMI(MBB, MI, DL, TII.get(MSP430::JMP)).addExternalSymbol(RestorerName);
+    return true;
+  }
 
   for (unsigned i = 0, e = CSI.size(); i != e; ++i)
     BuildMI(MBB, MI, DL, TII.get(MSP430::POP16r), CSI[i].getReg());
--- a/llvm/lib/Target/MSP430/MSP430FrameLowering.h
+++ b/llvm/lib/Target/MSP430/MSP430FrameLowering.h
@@ -38,6 +38,7 @@ public:
                                  MachineBasicBlock::iterator MI,
                                  ArrayRef<CalleeSavedInfo> CSI,
                                  const TargetRegisterInfo *TRI) const override;
+  const char *tailCallNameForRestoringCalleSavedRegisters(MutableArrayRef<CalleeSavedInfo> CSI) const;
   bool
   restoreCalleeSavedRegisters(MachineBasicBlock &MBB,
                               MachineBasicBlock::iterator MI,
--- a/llvm/test/CodeGen/MSP430/epilog-spilled-restore.ll
+++ b/llvm/test/CodeGen/MSP430/epilog-spilled-restore.ll
@@ -1,0 +1,14 @@
+; RUN: llc < %s | FileCheck %s
+
+target datalayout = "e-m:e-p:16:16-i32:16-i64:16-f32:16-f64:16-a:8-n8:16-S16"
+target triple = "msp430"
+
+; TODO check other counts
+; TODO check 7 regs! (something strange happens...)
+
+define void @pops_4_regs() nounwind {
+; CHECK-LABEL: pops_4_regs
+  call void asm sideeffect "", "~{r7},~{r8},~{r9},~{r10}"()
+  ret void
+; CHECK: jmp __mspabi_func_epilog_4
+}


atrosinenko:
On linkability, I expect no need to implement an automatic integration test, since any large enough code compiled as part of the nightly tests should catch this.

llvm/lib/Target/MSP430/MSP430FrameLowering.cpp
```diff
@@ -200,6 +200,31 @@ bool MSP430FrameLowering::spillCalleeSavedRegisters(
...
+    nullptr,
+    "__mspabi_func_epilog_1",
+    "__mspabi_func_epilog_2",
+    "__mspabi_func_epilog_3",
+    "__mspabi_func_epilog_4",
+    "__mspabi_func_epilog_5",
+    "__mspabi_func_epilog_6",
+    "__mspabi_func_epilog_7",
+  };
+  const unsigned TotalCount = sizeof(Registers) / sizeof(Registers[0]);
```
asl, line 218: there is array_lengthof in ADT/STLExtras.h

llvm/lib/Target/MSP430/MSP430FrameLowering.cpp
```diff
@@ -200,6 +200,31 @@ bool MSP430FrameLowering::spillCalleeSavedRegisters(
   return true;
 }
 
+const char *MSP430FrameLowering::tailCallNameForRestoringCalleSavedRegisters(MutableArrayRef<CalleeSavedInfo> CSI) const {
+  static unsigned Registers[] = {
+    /*MSP430::R4, ??? */ MSP430::R5, MSP430::R6, MSP430::R7,
+    MSP430::R8, MSP430::R9, MSP430::R10,
+  };
+  static const char *RestorerNames[] = {
```
asl, line 208: Can you provide the assembler code in the comments to illustrate the set of labels?

llvm/lib/Target/MSP430/MSP430FrameLowering.cpp
```diff
@@ -200,6 +200,31 @@ bool MSP430FrameLowering::spillCalleeSavedRegisters(
...
+    "__mspabi_func_epilog_3",
+    "__mspabi_func_epilog_4",
+    "__mspabi_func_epilog_5",
+    "__mspabi_func_epilog_6",
+    "__mspabi_func_epilog_7",
+  };
+  const unsigned TotalCount = sizeof(Registers) / sizeof(Registers[0]);
+  const unsigned PoppedCount = CSI.size();
+  for (unsigned i = 0; i < PoppedCount; ++i) {
+    if (CSI[i].getReg() != Registers[TotalCount - PoppedCount + i]) {
```
asl, line 221: I think we'd need to assert here to ensure that TotalCount >= PoppedCount

llvm/lib/Target/MSP430/MSP430FrameLowering.cpp
```diff
@@ -212,6 +237,12 @@ bool MSP430FrameLowering::restoreCalleeSavedRegisters(
   MachineFunction &MF = *MBB.getParent();
   const TargetInstrInfo &TII = *MF.getSubtarget().getInstrInfo();
 
+  const char *RestorerName = tailCallNameForRestoringCalleSavedRegisters(CSI);
```
asl, line 240: We have problem here. The RET after JUMP is still on place and nothing will erase it. Or maybe the branch folder does this already for us?

llvm/lib/Target/MSP430/MSP430FrameLowering.cpp
atrosinenko: Do you mean prepending this array with comment lines containing plain assembler code or "inlining" these comments right after label names?

llvm/lib/Target/MSP430/MSP430FrameLowering.cpp
atrosinenko: > We have problem here. The RET after JUMP is still on place and nothing will erase it.

Exactly. This is what I TODO'ed as "get rid of the dangling ret at the very end of a function" (strictly speaking, not only function but every terminating BB, if I get it right). Right now I'm trying to drop it, but just calling

```cpp
MBB.getLastNonDebugInstr()->removeFromParent();
```

generates an assertion slightly later:

```
Command Output (stderr):
--
Can only insert epilog into returning blocks
UNREACHABLE executed at /home/ast/projects/llvm-project/llvm/lib/Target/MSP430/MSP430FrameLowering.cpp:121!
...
 #7 0x000056363ad77aa9 bindingsErrorHandler(void*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) /home/ast/projects/llvm-project/llvm/lib/Support/ErrorHandling.cpp:219:55
 #8 0x0000563639470068 llvm::MSP430FrameLowering::emitEpilogue(llvm::MachineFunction&, llvm::MachineBasicBlock&) const /home/ast/projects/llvm-project/llvm/lib/Target/MSP430/MSP430FrameLowering.cpp:119:22
 #9 0x000056363a0c186b (anonymous namespace)::PEI::insertPrologEpilogCode(llvm::MachineFunction&) /home/ast/projects/llvm-project/llvm/lib/CodeGen/PrologEpilogInserter.cpp:1100:3
#10 0x000056363a0bdc71 (anonymous namespace)::PEI::runOnMachineFunction(llvm::MachineFunction&) /home/ast/projects/llvm-project/llvm/lib/CodeGen/PrologEpilogInserter.cpp:259:18
```

Another thing that seems fragile to me (even if It would be appropriate to refactor that code to check *before* emitting epilog) is that I'm not sure non-lazy instruction removal would not invalidate some iterators up the stack. Is there some delayed removal machinery like `setIsDead()` on operand?

llvm/lib/Target/MSP430/MSP430FrameLowering.cpp
asl: Prepending (or, rather, appending) looks ok. Something like:
```
"__mspabi_func_epilog_7", // __mspabi_func_epilog_7: POP R4
"__mspabi_func_epilog_6", //__mspabi_func_epilog_6: POP R5
```

llvm/lib/Target/MSP430/MSP430FrameLowering.cpp
asl: How about the following: introduce the special pseudo-instruction that would be terminator. Let's call it EPILOGUE. It would have single imm argument – the number of registers. Emit the instruction here. And just during the pseudo instruction expansion erase the RET that would be after it.

llvm/lib/Target/MSP430/MSP430FrameLowering.cpp
atrosinenko: Another thing to consider: after that change, I would have two arrays: `Registers` and `RestorerNames` that seem to be logically aligned with each other, so it seems reasonable to convert them to a single array of `struct`s or so. In other projects I have seen using this together with parameterized `#define`s. Here it seems quite handy to define something like

```cpp
struct RestorerNameMapping {
  const char *name;
  unsigned reg;
};
static RestorerMapping Restorers[] = {
#define DEF(name_suff, reg) { \
    .name = "__mspabi_func_epilog_" name_suff, \
    .reg = reg \
  }
DEF("7", MSP430::R4),
DEF("6", MSP430::R5),
DEF("5", MSP430::R6),
DEF("4", MSP430::R7),
DEF("3", MSP430::R8),
DEF("2", MSP430::R9),
DEF("1", MSP430::R10),
#undef DEF
};
```

On one hand, this can make it harder to read. On the other hand, this eliminates some pitfalls:
* no way to mis-spell _one of_ `__mspabi_func_epilog` prefixes while retaining these strings in the form of literal constants (so no heap allocation or possibly dangling pointers to stack-allocated buffers)
* the two arrays have the same size by construction (because now these are a single array)
* after the `DEF` macro is ultimately deciphered by the reader, it should be much simpler to review this code due to less boilerplate

atrosinenko: Oops, this is C++ after all, so no `#define`s even required...

cr1901:
>ultimately improve it to upstreamable quality

I would like to test this against Rust ASAP. Do you suggest I wait until its upstreamable quality or start running smoke tests as soon as the feature seems to be working?

llvm/lib/Target/MSP430/MSP430EpilogueOptimization.cpp
```diff
@@ -0,0 +1,142 @@
...
+    InstrToRemove->eraseFromParent();
+  }
+}
+
+bool MSP430EpilogueOptimization::tryReplacingEpilogue(MachineBasicBlock &MBB) {
+  MachineInstr *LastMatchedPopInstr = nullptr;
+  unsigned MatchedRegisters = 0;
+
+  MachineBasicBlock::iterator Terminator = MBB.getFirstTerminator();
+  if (Terminator == MBB.end() || Terminator->getOpcode() != MSP430::RET) {
```
asl, line 100: Typically LLVM style guide suggests not to use braces on 1 line bodies unless it would be ambiguous. Same below

llvm/lib/Target/MSP430/MSP430EpilogueOptimization.cpp
```diff
@@ -0,0 +1,142 @@
...
+    return false;
+  }
+
+  // Start with one instruction *before* RET and go backwards
+  for (auto it = std::next(Terminator.getReverse()), e = MBB.rend();
```
asl, line 105: Usually we're using uppercase for variables, e.g. auto I = std::next(), E = MBB.rend()

llvm/lib/Target/MSP430/MSP430EpilogueOptimization.cpp
```diff
@@ -0,0 +1,142 @@
...
+       it != e && MatchedRegisters < array_lengthof(RestoredRegisters); ++it) {
+    if (it->isDebugInstr()) {
+      continue;
+    }
+
+    if (it->getOpcode() != MSP430::POP16r ||
+        it->getOperand(0).getReg() != RestoredRegisters[MatchedRegisters]) {
+      break;
+    } else {
```
asl, line 114: No need to else here, since there is a break before :)

llvm/lib/Target/MSP430/MSP430EpilogueOptimization.cpp
```diff
@@ -0,0 +1,142 @@
...
+  const char *LibCallName = EpilogueLibCallNames[MatchedRegisters];
+  replaceTailWithJmpToLibCall(MBB, *LastMatchedPopInstr, LibCallName);
+
+  return true;
+}
+
+bool MSP430EpilogueOptimization::runOnMachineFunction(MachineFunction &MF) {
+  TII = static_cast<const MSP430InstrInfo *>(MF.getSubtarget().getInstrInfo());
+
+  bool wasModified = false;
```
asl, line 133: Should be uppercase. E.g. simple ``Modified``

atrosinenko:
@cr1901 

> I would like to test this against Rust ASAP. Do you suggest I wait until its upstreamable quality or start running smoke tests as soon as the feature seems to be working?

I expect this feature to be testable already. I'm not sure it would not have unresolved symbols during linking, but this depends on the sysroot being used for cross-compilation anyway. *Current* implementation is not expected to break something *silently* - just link-time errors :)

On the other hand, I'm not absolutely sure this patch will not be rewritten from scratch once again.

atrosinenko:
* Squashed commits and updated for #12 (renaming `FP` -> `R4`).
* Made tests slightly stricter

atrosinenko:
When compiling `newlib` with this patch applied, code generator crashes several times. The crash can be reproduced with

```ll
target datalayout = "e-m:e-p:16:16-i32:16-i64:16-f32:16-f64:16-a:8-n8:16-S16"
target triple = "msp430"

@x = external dso_local global i16

define dso_local void @_dtoa_r(i16*, i64) {
; First, make the function not fit into about 1024 bytes
  store volatile i16 1, i16* @x
  store volatile i16 1, i16* @x
; ... 200-300 lines here ...
  store volatile i16 1, i16* @x
  store volatile i16 1, i16* @x
; Then force emitting jump to LibCall
  call void asm sideeffect "", "~{r4},~{r5},~{r6},~{r7},~{r8},~{r9},~{r10}"()
  ret void
}
```

```
$ /path/to/bin/llc < test.ll 
        .text
        .file   "<stdin>"
llc: /old_ssd/ast/msp430-projects/llvm-project/llvm/include/llvm/CodeGen/MachineOperand.h:551: llvm::MachineBasicBlock* llvm::MachineOperand::getMBB() const: Assertion `isMBB() && "Wrong MachineOperand accessor"' failed.
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace.
Stack dump:
0.      Program arguments: /old_ssd/ast/msp430-projects/llvm-project/build-rel-assert/bin/llc 
1.      Running pass 'Function Pass Manager' on module '<stdin>'.
2.      Running pass 'MSP430 Branch Selector' on function '@_dtoa_r'
 #0 0x00005614194ca9ce llvm::sys::PrintStackTrace(llvm::raw_ostream&) /old_ssd/ast/msp430-projects/llvm-project/llvm/lib/Support/Unix/Signals.inc:568:3
 #1 0x00005614194c8744 llvm::sys::RunSignalHandlers() /old_ssd/ast/msp430-projects/llvm-project/llvm/lib/Support/Signals.cpp:68:20
 #2 0x00005614194c8e85 SignalHandler(int) /old_ssd/ast/msp430-projects/llvm-project/llvm/lib/Support/Unix/Signals.inc:396:31
 #3 0x00007f1966743540 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x15540)
 #4 0x00007f19661d83eb raise /build/glibc-t7JzpG/glibc-2.30/signal/../sysdeps/unix/sysv/linux/raise.c:51:1
 #5 0x00007f19661b7899 abort /build/glibc-t7JzpG/glibc-2.30/stdlib/abort.c:81:7
 #6 0x00007f19661b7769 get_sysdep_segment_value /build/glibc-t7JzpG/glibc-2.30/intl/loadmsgcat.c:509:8
 #7 0x00007f19661b7769 _nl_load_domain /build/glibc-t7JzpG/glibc-2.30/intl/loadmsgcat.c:970:34
 #8 0x00007f19661c9006 (/lib/x86_64-linux-gnu/libc.so.6+0x37006)
 #9 0x000056141824981a llvm::MCInstrInfo::get(unsigned int) const /old_ssd/ast/msp430-projects/llvm-project/llvm/include/llvm/MC/MCInstrInfo.h:61:5
#10 0x000056141824981a llvm::MCInstrInfo::get(unsigned int) const /old_ssd/ast/msp430-projects/llvm-project/llvm/include/llvm/MC/MCInstrInfo.h:60:22
#11 0x000056141824981a expandBranches /old_ssd/ast/msp430-projects/llvm-project/llvm/lib/Target/MSP430/MSP430BranchSelector.cpp:196:57
#12 0x000056141824981a runOnMachineFunction /old_ssd/ast/msp430-projects/llvm-project/llvm/lib/Target/MSP430/MSP430BranchSelector.cpp:248:24
#13 0x000056141824981a (anonymous namespace)::MSP430BSel::runOnMachineFunction(llvm::MachineFunction&) /old_ssd/ast/msp430-projects/llvm-project/llvm/lib/Target/MSP430/MSP430BranchSelector.cpp:224:6
#14 0x00005614189509f9 llvm::MachineFunctionPass::runOnFunction(llvm::Function&) /old_ssd/ast/msp430-projects/llvm-project/llvm/lib/CodeGen/MachineFunctionPass.cpp:73:33
#15 0x00005614189509f9 llvm::MachineFunctionPass::runOnFunction(llvm::Function&) /old_ssd/ast/msp430-projects/llvm-project/llvm/lib/CodeGen/MachineFunctionPass.cpp:38:6
#16 0x0000561418d63d43 llvm::FPPassManager::runOnFunction(llvm::Function&) /old_ssd/ast/msp430-projects/llvm-project/llvm/lib/IR/LegacyPassManager.cpp:1483:7
#17 0x0000561418d647b9 llvm::ilist_node_impl<llvm::ilist_detail::node_options<llvm::Function, true, false, void> >::getNext() /old_ssd/ast/msp430-projects/llvm-project/llvm/include/llvm/ADT/ilist_node.h:66:66
#18 0x0000561418d647b9 llvm::ilist_iterator<llvm::ilist_detail::node_options<llvm::Function, true, false, void>, false, false>::operator++() /old_ssd/ast/msp430-projects/llvm-project/llvm/include/llvm/ADT/ilist_iterator.h:157:25
#19 0x0000561418d647b9 llvm::FPPassManager::runOnModule(llvm::Module&) /old_ssd/ast/msp430-projects/llvm-project/llvm/lib/IR/LegacyPassManager.cpp:1517:22
#20 0x0000561418d64b73 runOnModule /old_ssd/ast/msp430-projects/llvm-project/llvm/lib/IR/LegacyPassManager.cpp:1584:7
#21 0x0000561418d64b73 llvm::legacy::PassManagerImpl::run(llvm::Module&) /old_ssd/ast/msp430-projects/llvm-project/llvm/lib/IR/LegacyPassManager.cpp:1695:55
#22 0x00005614181cc50b compileModule /old_ssd/ast/msp430-projects/llvm-project/llvm/tools/llc/llc.cpp:627:66
#23 0x00005614181cc50b main /old_ssd/ast/msp430-projects/llvm-project/llvm/tools/llc/llc.cpp:360:35
#24 0x00007f19661b91e3 __libc_start_main /build/glibc-t7JzpG/glibc-2.30/csu/../csu/libc-start.c:342:3
#25 0x0000561418219e8e _start (/old_ssd/ast/msp430-projects/llvm-project/build-rel-assert/bin/llc+0x41be8e)
Aborted
```
`MSP430BSel::expandBranches` tries to extract JMP target as a `MachineBasicBlock` when it is LibCall:

```cpp
  MachineBasicBlock *DestBB = MI->getOperand(0).getMBB();
```

The issues itself seems not too hard to fix. But it would be useful to add a test on this. The problem is due to condition inside the `MSP430BSel::runOnMachineFunction` function:

```cpp
  unsigned FunctionSize = measureFunction(BlockOffsets);
  // If the entire function is smaller than the displacement of a branch field,
  // we know we don't need to expand any branches in this
  // function. This is a common case.
  if (isInRage(FunctionSize)) {
    return false;
  }
```

My reproducer triggers this issue well _now_, but my general concern is "how to handle such simple and generally useful heuristics that can make some testcase no-op" (such as if the test would be implemented just as a couple of instructions _before_ this optimization was added)?

atrosinenko:
Hmm... Can we emit these epilogues via `jmp external_symbol` at all? Looks like we have no information on its displacement before actual linking.

asl:
The issue here is that JMP instruction expects basic block. We just need another variant with different opcode that expects external symbol as an argument, so the code in branch expansion will not touch it.

atrosinenko:
> The issue here is that JMP instruction expects basic block.

Is it because its input list is described as `(ins jmptarget:$dst)` in `.td`?

If I get it right, `MSP430::Bi` does exactly what is required.

asl:
> Is it because its input list is described as (ins jmptarget:$dst) in .td?

Right

> If I get it right, MSP430::Bi does exactly what is required.

I'm not 100% sure. We'd need long jump, not PC-relative branch, please double check.

atrosinenko:
Replaced `JMP` with `Bi`. Now epilogues are generated like the following:
```
      76: 30 40 00 00                   br      #0
                        00000078:  R_MSP430_16_BYTE     __mspabi_func_epilog_7
```

The fact that epilogue is emitted via `br #__mspabi_func_epilog_N` is checked by regression tests. Hope this ensures jumps are always "long-enough". I'm not familiar with precise MSP430 assembler mnemonics, though.

On the other hand, now epilogues are always 4-byte long. Still, this never makes code larger - just not as small as it could be.

llvm/lib/Target/MSP430/MSP430EpilogueOptimization.cpp
```diff
@@ -0,0 +1,146 @@
...
+    MI = MI->getNextNode();
+    InstrToRemove->eraseFromParent();
+  }
+}
+
+bool MSP430EpilogueOptimization::tryReplacingEpilogue(MachineBasicBlock &MBB) {
+  MachineInstr *LastMatchedPopInstr = nullptr;
+  unsigned MatchedRegisters = 0;
+
+  MachineBasicBlock::iterator Terminator = MBB.getFirstTerminator();
```
asl, line 104: Ok, after thinking about this a bit more I believe we have an issue here. It's perfectly ok for a basic block to contain multiple terminators. For example, it could be a conditional branch and then return, so taking first terminator would not work here. I think you could craft a testcase with a simple do {} while() loop and return afterwards.

atrosinenko: Agree, I have tested with something like
```ll
define i16 @f() noinline optnone {
entry:
  br label %do.cond

do.cond:
  br i1 1, label %do.cond, label %do.end

do.end:
  ret i16 42
}
```
being converted to
```
$ /path/to/llc -stop-after=legalizer < while.ll
// ...
body:             |
  bb.0.entry:
    successors: %bb.1(0x80000000)
  
    JMP %bb.1
  
  bb.1.do.cond:
    successors: %bb.1(0x40000000), %bb.2(0x40000000)
  
    $r12 = MOV16rc 0
    CMP16rc killed $r12, 1, implicit-def $sr
    JCC %bb.1, 1, implicit $sr
    JMP %bb.2
  
  bb.2.do.end:
    $r12 = MOV16ri 42
    RET implicit $r12
```
Not yet sure how to trick code generator into emitting conditional jump and return inside a single MBB, but this is definitely worth fixing.

asl: No need to trick. Use llc --print-after to dump MIR. You could edit it as you wish and run codegen on it :)

Probably in your case 'ret void' will work ;) Though... no additional code is allowed between the terminators (unless the code is predicated), so we cannot restore regs there. So, maybe the approach will still work, yes, though is a bit fragile... I'd scan from the end of BB for the sake of clarity.

atrosinenko: So, maybe it is worth leaving it as is, just convert the last for-loop in the `replaceTailWithBrToLibCall`
```cpp
for (MachineInstr *MI = &FirstInstrToReplace; MI != nullptr;) { ...
```
to remove till `FirstInstrToReplace`, *inclusive*, instead of removing till the end. And add an explanation. This way, the modification would produce identical behavior and leave MBB not less well-formed than it was before :)

atrosinenko: I have pushed a fixed version but without an explanation yet. Is there some documentation on "no additional code is allowed between the terminators" and other such things? Looks like it is less documented than the LLVM IR.

atrosinenko:
Sent for public discussion: https://reviews.llvm.org/D84397

Issue #10, opened 2020-04-21
[RFC] Validate inline asm constraints in llc


atrosinenko:
While writing a regression test for #8 I have spent some time trying to understand why I cannot clobber `r4` via `__asm__` while clobbering `r5` - `r10` works fine. The code was like this:

```ll
define i16 @pops_7_regs() nounwind {
; CHECK-LABEL: pops_7_regs
  call void asm sideeffect "", "~{r4},~{r5},~{r6},~{r7},~{r8},~{r9},~{r10}"()
  ret i16 42
; CHECK:     jmp __mspabi_func_epilog_7
; CHECK-NOT: ret
; CHECK:     -- End function
}
```

`llc` was silently ignoring `~{r4}` constraint, then generating otherwise correct epilogue (for `r5` - `r10` being clobbered) and replacing it with `jmp __mspabi_func_epilog_6`. The fix was literally two-byte-size: replace `r4` register name with `fp`.

While `llc` is internal tool for LLVM developers, it still may be worth implement some checks where it is trivial to do.

For both `MSP430` and `X86`, the register names are checked by `clang` and are NOT checked by `llc`, so this is consistent behavior.

The other problem with particularly the MSP430 backend was that `clang` does accept `r4` as a valid register name passing it through as `r4`. So, now it is a user-facing issue, as well.

My proposal would be to implement some assertions, at first, where it is most trivial and makes any sense. IMO, these should be specifically **assertions** and not some neat error messages such as well-known "undefined symbol" errors, otherwise it would implicitly suggest we have full constraint validation, which would be not true, at least for the first time.

PR #11, opened 2020-04-21
[WIP] [CodeGen] Validate clobber constraint for inline asm

atrosinenko:
This is an early draft fix of #10 w.r.t. clobber constraints (they seem to be quite trivial according to the [LLVM IR reference](https://llvm.org/docs/LangRef.html#clobber-constraints)). This PR **does not** cover any other constraint kinds.

It **does not** pass regression tests now. While it may indicate plainly incorrect implementation of this PR (and most probably this is the case), sometimes it triggers on suspicious parts of code. For example, `CodeGen/X86/swifterror.ll` fails when run as

```
llc -verify-machineinstrs -mtriple=i386-apple-darwin -disable-block-placement < swifterror.ll
```

due to usage of

```ll
  call void asm sideeffect "nop", "~{r12}"()
```
for testing on `i386`. Whether this is an error or not, I don't know. Still, this looks slightly suspicious.

Please note this PR is most probably very broken right now and is mostly here for discussion on what to check and what to ignore (and hopefully to produce some correct validation someday).

### DIFF
--- a/llvm/lib/CodeGen/SelectionDAG/FunctionLoweringInfo.cpp
+++ b/llvm/lib/CodeGen/SelectionDAG/FunctionLoweringInfo.cpp
@@ -209,6 +209,15 @@ void FunctionLoweringInfo::set(const Function &fn, MachineFunction &mf,
               std::pair<unsigned, const TargetRegisterClass *> PhysReg =
                   TLI->getRegForInlineAsmConstraint(TRI, Op.ConstraintCode,
                                                     Op.ConstraintVT);
+              bool CanParseMachineRegisterOrMemory =
+                  PhysReg.first != 0 ||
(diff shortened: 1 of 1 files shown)


atrosinenko:
Another random observation: [llvm/test/CodeGen/AVR/no-print-operand-twice.ll](https://github.com/access-softek/llvm-project/blob/d33c7de8e11f2727ef5e325d931ae94af8619bf9/llvm/test/CodeGen/AVR/no-print-operand-twice.ll#L6) contains a line

```ll
  tail call void asm sideeffect "/* result: ${0:c} */", "i,~{dirflag},~{fpsr},~{flags}"( i64 68719476738 )
```

This cleanly looks like an AVR assembly converted from an X86 one. This particular case is most probably harmless, though.

PR #17, opened 2020-04-29
[MSP430] Update toolchain description for msp430-gcc-8.3.1.25

atrosinenko:
This patch updates MSP430 toolchain description for current mspgcc package provided by TI. It does not try to be backward-compatible yet.

This patch expects the check for `options::OPT_fexceptions` to always return correct value based on `-fexceptions`, `-fno-exceptions` and the default value, that is probably not the case...

Actual object and archive file sequence is _based_ on "deciphered" output of `msp430-elf-gcc -dumpspecs` and `gcc/gcc/config/msp430/msp430.h` source file. Still, it lacks some parts such as stack-smashing protector that seems to be supported on MSP430 according to presence of `libssp.a`. Of course, it omits all the sanitizer runtimes.

Provided `uart-test.c` contains the UART example [from MSP430-Emulator documentation](http://www.msp430emulator.com/examples.html#2), both commands

```sh
/path/to/bin/clang -target msp430 --sysroot=/path/to/msp430gcc uart-test.c -v -mmcu=msp430g2553 -o uart-test
/path/to/bin/clang -target msp430 --sysroot=/path/to/msp430gcc uart-test.c -v -mmcu=msp430g2553 -o uart-test -fexceptions
```

produce executables that can be run on emulator by
* Uploading firmware
* `dump 0xfffe`
* `set pc 0x<value of reset vector>`
* `run`

Here, the `--sysroot` option points to TI's binary distribution:
```
$ cd /path/to/msp430gcc
 bin                 lib                        Revisions_MSPDebugStack.txt
 common              libexec                    rollbackBackupDirectory
 docs                msp430.dat                 share
 examples            msp430-elf                 Software_License_Agreement.pdf
'GPLv3 Notice.txt'   MSP430-GCC_manifest.html   uninstall
 include            'Release Notes.txt'         uninstall.dat
 install_scripts     Revisions_Header.txt       version.properties
$ ls include/
cc430f5123.h             msp430f5308.ld                    msp430fr2633.ld
cc430f5123.ld            msp430f5308_symbols.ld            msp430fr2633_symbols.ld
cc430f5123_symbols.ld    msp430f5309.h                     msp430fr2672.h
cc430f5125.h             msp430f5309.ld                    msp430fr2672.ld
$ ls msp430-elf/
bin  include  lib
...

```

### DIFF
--- a/clang/lib/Driver/ToolChains/Gnu.cpp
+++ b/clang/lib/Driver/ToolChains/Gnu.cpp
@@ -1503,7 +1503,9 @@ static bool findMSP430Multilibs(const Driver &D,
                                 StringRef Path, const ArgList &Args,
                                 DetectedMultilibs &Result) {
   FilterNonExistent NonExistent(Path, "/crtbegin.o", D.getVFS());
-  Multilib MSP430Multilib = makeMultilib("/430");
+  Multilib MSP430Multilib = Args.hasArg(options::OPT_fexceptions) ?
+        makeMultilib("/430/exceptions") : makeMultilib("/430");
+
   // FIXME: when clang starts to support msp430x ISA additional logic
   // to select between multilib must be implemented
   // Multilib MSP430xMultilib = makeMultilib("/large");
--- a/clang/lib/Driver/ToolChains/MSP430.cpp
+++ b/clang/lib/Driver/ToolChains/MSP430.cpp
@@ -128,7 +128,7 @@ MSP430ToolChain::MSP430ToolChain(const Driver &D, const llvm::Triple &Triple,
   }
 
   SmallString<128> SysRootDir(computeSysRoot());
-  llvm::sys::path::append(SysRootDir, "lib", MultilibSuf);
+  llvm::sys::path::append(SysRootDir, "msp430-elf", "lib", MultilibSuf);
   addPathIfExists(D, SysRootDir, getFilePaths());
 }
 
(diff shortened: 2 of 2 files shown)


asl:
Please also updated the driver tests and submit as a separate PR.

atrosinenko:
As I understood, `OPT_fexception` handling needs further rewriting as well: in [clang/lib/Driver/ToolChains/Clang.cpp](https://github.com/access-softek/llvm-project/blob/525a204df258915730087f3444c7c002d946a7a4/clang/lib/Driver/ToolChains/Clang.cpp#L415), there exists some non-trivial code to compute whether exceptions are enabled or not. For MSP430 backend, we have probably to use at least 

```cpp
  Args.hasFlag(options::OPT_fexceptions, options::OPT_fno_exceptions, false);
```
or better explicitly expressing the default state...

On the other hand, in GCC `spec`, this is described as `%{fexceptions:crtend.o}%{!fexceptions:crtend_no_eh.o}`, so my implementation is even correctly aligned with GCC one provided it has no special magic w.r.t. `-fsomething`/`-fno-something` handling built in.

Meanwhile, in [clang/lib/Driver/ToolChains/Fuchsia.cpp](https://github.com/access-softek/llvm-project/blob/525a204df258915730087f3444c7c002d946a7a4/clang/lib/Driver/ToolChains/Fuchsia.cpp#L192), there exists nicer multilib description for roughly the same case.

asl:
Right. But given that this is C++ stuff, please just fill an issue, we will take care of it with lower priority :)

atrosinenko:
For the record, switching to `*_no_eh.o` versions by default (when no `-fexceptions` is explicitly given) should slightly decrease code size, so this PR seems to be not only "making it automatically discover newlib" :)

Now, I will update clang driver test, so what is the policy on supported sysroot versions? Are we planning choosing (or even autodetecting) TI's sysroot version to be used - now or in the near future?

asl:
We need to test whatever is implemented

PR #20, opened 2020-05-06
[MSP430] Toolchain description updates for clang-rt

atrosinenko:
In preparation for compiler-rt built-ins support on MSP430, push less questionable patch first, so my other experimental branches can be rebased onto it.


clang/lib/Driver/ToolChains/MSP430.cpp
```diff
@@ -161,6 +161,7 @@ void MSP430ToolChain::addClangTargetOptions(const ArgList &DriverArgs,
                                             ArgStringList &CC1Args,
                                             Action::OffloadKind) const {
   CC1Args.push_back("-nostdsysteminc");
+  CC1Args.push_back("-D__ELF__");
```
asl, line 164: If this is necessary, then it should be properly presented as a builtin define.

atrosinenko: Fixed.

PR #21, opened 2020-05-06
[MSP430] Initial fixes for built-ins on MSP430

atrosinenko:
Here are mostly guessed fixes to `compiler-rt` that make it compile for MSP430 target. These are definitely WIP at the time of writing.

It can be compiled with something like
```sh
llvm_build=/path/to/llvm/build-directory
sysroot=/path/to/msp430gcc

cflags="-target msp430 -mmcu=msp430g2553 --gcc-toolchain=$sysroot -I$sysroot/include -L$sysroot/include"

cmake "$@" ../compiler-rt -G Ninja \
    -DCOMPILER_RT_DEFAULT_TARGET_TRIPLE="msp430-none-elf" \
    -DCOMPILER_RT_EMULATOR="$sysroot/bin/msp430-elf-run" \
    -DCOMPILER_RT_INCLUDE_TESTS=ON \
    -DCOMPILER_RT_BUILD_BUILTINS=ON \
    -DCMAKE_C_COMPILER=$llvm_build/bin/clang \
    -DCMAKE_CXX_COMPILER=$llvm_build/bin/clang++ \
    -DCMAKE_AR=$llvm_build/bin/llvm-ar \
    -DCMAKE_NM=$llvm_build/bin/llvm-nm \
    -DCMAKE_RANLIB=$llvm_build/bin/llvm-ranlib \
    -DLLVM_CONFIG_PATH=$llvm_build/bin/llvm-config \
    -DCMAKE_C_FLAGS="$cflags" \
    -DCMAKE_CXX_FLAGS="$cflags" \
    -DCMAKE_ASM_FLAGS="$cflags" \
    -DCOMPILER_RT_BAREMETAL_BUILD=ON
```

Prior to #20 the `__ELF__` macro have to be set as well.


atrosinenko:
I have managed to run most of compiler-rt unit tests via `msp430-elf-run` simulator (shipped with gdb, see updated PR description) when I forgot dropping some `-lgcc` from linker command line. :) Without `-lgcc`, there are obviously lots of unresolved references to `__mspabi_XYZ`. According to `MSP430TargetLowering`, most of them are just renamed `libgcc` symbols, so what is the preferred way to map these names?

asl:
Create an alias? Similar to ARM EABI functions?

atrosinenko:
Yes, I expected something similar to other targets. As I understand, sometimes it is defined just like the following:

https://github.com/access-softek/llvm-project/blob/40ae542656c264727049f14eb629de08d2cee09a/compiler-rt/lib/builtins/muldi3.c#L38-L51

at the end of common `.c` file (not taking into account native asm implementations naturally residing under 'builtins/arm/'). I would prefer to describe all these aliases in a single `.c` or `.S` file under `builtins/msp430` instead, but this is probably against the policy of separate files for separate functions. On the other hand, spreading such tiny aliases across lots of files may be quite hard to review and may be not very maintainable.

Ideally, I expect some autogenerated file that would be aligned with `MSP430TargetLowering` by construction (still, I'm not sure it is possible in general case).

asl:
Just put into separate files :)

atrosinenko:
_Just to not perform massive rewriting afterwards :)_ Do I get it right, separate files `builtins/msp430/__mspabi_XYZ.S`, each containing a standard copyright header and a couple of macroses aliasing it to existing generic implementation for now? (Except for some msp-specific routines, of course)

asl:
Ok. In msp430 directory

atrosinenko:
The problem with aliasing via `msp430/*.S` was that I cannot alias a symbol from another object file (this can be done, but with linker script, not just another `mspabi-aliases.o`...).

On the other hand, it seems not too handy to spread aliases for every new target across `.c` files with corresponding generic implementations. What if just add an `ALIAS<defined function name>` macro "invocation" for every libcall? Then only a single "generic" header needs an adjustment for every new target, while all target-specific aliases can reside in a single header, not cluttering every generic `.c` file.

asl:
Sorry, I do not follow.

atrosinenko:
While googling how to emit some symbol as an alias to another one, I have found some suggestions on using `.set to_symbol, from_symbol`. But, as I understood, there exist some generic issues for referencing a symbol not defined in the same object file (relocations don't work for that cases, at least generically, they say...) and wasting 4*renamed_symbol_count bytes for this (`br original_symbol`) may be not very good for such low-memory target.

atrosinenko:
So, there exist other possibilities: either the most trivial one - use assignments via a linker script or alias in the correct translation unit. Linker script may require some non-trivial building. On the other hand, aliasing in the generic C-sources are already used for ARM target. My proposed approach just make it possible to localize changes for new target support in a couple of files instead of spreading them over at most O(libcall count) files.

PS:
1. *My* patch still have to touch every required libcall source file
2. In the latest commit of this PR, there is only a single file manually patched as an example
3. ARM aliases may probably be refactored that way if appropriate

asl:
We cannot use linker script in general. It may be custom or not used at all. So, may I suggest us to go simple for now and mimic what ARM EABI is doing?

atrosinenko:
Current `check-builtins` status:

```
Failing Tests (36):
  Builtins-msp430-linux :: absvdi2_test.c
  Builtins-msp430-linux :: absvsi2_test.c
  Builtins-msp430-linux :: addtf3_test.c
  Builtins-msp430-linux :: ashldi3_test.c
  Builtins-msp430-linux :: ashrdi3_test.c
  Builtins-msp430-linux :: clzsi2_test.c
  Builtins-msp430-linux :: compiler_rt_logb_test.c
  Builtins-msp430-linux :: compiler_rt_logbf_test.c
  Builtins-msp430-linux :: ctzsi2_test.c
  Builtins-msp430-linux :: divdc3_test.c
  Builtins-msp430-linux :: divdf3_test.c
  Builtins-msp430-linux :: divsc3_test.c
  Builtins-msp430-linux :: divsf3_test.c
  Builtins-msp430-linux :: divtc3_test.c
  Builtins-msp430-linux :: fixdfdi_test.c
  Builtins-msp430-linux :: fixsfdi_test.c
  Builtins-msp430-linux :: fixunsdfdi_test.c
  Builtins-msp430-linux :: fixunssfdi_test.c
  Builtins-msp430-linux :: floatdidf_test.c
  Builtins-msp430-linux :: floatdisf_test.c
  Builtins-msp430-linux :: floatundidf_test.c
  Builtins-msp430-linux :: floatundisf_test.c
  Builtins-msp430-linux :: lshrdi3_test.c
  Builtins-msp430-linux :: muldc3_test.c
  Builtins-msp430-linux :: mulodi4_test.c
  Builtins-msp430-linux :: mulsc3_test.c
  Builtins-msp430-linux :: mulvdi3_test.c
  Builtins-msp430-linux :: paritydi2_test.c
  Builtins-msp430-linux :: paritysi2_test.c
  Builtins-msp430-linux :: popcountdi2_test.c
  Builtins-msp430-linux :: popcountsi2_test.c
  Builtins-msp430-linux :: powidf2_test.c
  Builtins-msp430-linux :: powisf2_test.c
  Builtins-msp430-linux :: subtf3_test.c
  Builtins-msp430-linux :: truncdfsf2_test.c
  Builtins-msp430-linux :: udivmoddi4_test.c


Testing Time: 2.04s
  Unsupported Tests  : 95
  Expected Passes    : 71
  Expected Failures  :  1
  Unexpected Failures: 36
```

asl:
I would ignore all FP tests for now and try to understand the integer ones.

atrosinenko:
Not surprisingly, my draft implementations for msp430-specific shift LibCalls were wrong :)

After they were fixed:
```
  Unsupported Tests  : 95
  Expected Passes    : 82
  Expected Failures  :  1
  Unexpected Failures: 31
```

Specifically, the following tests were fixed:
```
  Builtins-msp430-linux :: ashldi3_test.c
  Builtins-msp430-linux :: ashrdi3_test.c
  Builtins-msp430-linux :: clzsi2_test.c
  Builtins-msp430-linux :: ctzsi2_test.c
  Builtins-msp430-linux :: lshrdi3_test.c
```
and 6 other were added (for these 6 shift LibCalls).

atrosinenko:
> Shouldn't this be a part of COMPILER_RT_ABI?

According to MSP430 EABI, "6.3 Special Register Conventions for Helper Functions", there is only a small subset of `__mspabi_*` functions that operate on two 64-bit arguments - they use this optimization. So this is probably correct but slightly misleading: `CallingConv::MSP430_BUILTIN` is that "modified" convention and not the one used for *every* LibCall. It would probably be better to leave it as-is and add an explanation.

asl:
What will break if we'd use this CC for all builtins?

atrosinenko:
Right now, for example, I have adjusted `MSP430CallingConv.td` by adding
```
  CCIfCC<"CallingConv::MSP430_BUILTIN", CCPromoteToType<i64>>,
```
Still it is just guessed in many ways (but seems to ultimately work...), so there may exist more correct way to express this. I agree, using `MSP430_BUILTIN` for, well, all MSP430 builtins would make sense :)

But I'm not sure this can be trivially achieved without specially marking LibCalls that need such an adjustment. For example, part 6.3 lists "The following functions take 2 64-bit values". Among others, there is `__mspabi_srlll` function. But even in MSP430 EABI Table 10, it is declared as `uint64 __mspabi_srlll(uint64 x, int16 n);`, so one cannot just match on LibCall with two `i64` arguments. This still *may* be achievable but it would require further investigations.

atrosinenko:
```
  Unsupported Tests  : 95
  Expected Passes    : 94
  Expected Failures  :  1
  Unexpected Failures: 22
```

compiler-rt/lib/builtins/divhi3.c
```diff
@@ -0,0 +1,34 @@
...
+//
+// This file implements __divhi3 for the compiler_rt library.
+//
+//===----------------------------------------------------------------------===//
+
+#include "int_lib.h"
+
+// Returns: a / b
+
+COMPILER_RT_ABI hi_int __divhi3(hi_int a, hi_int b) {
```
asl, line 17: Can we have it msp430 specific for now?

compiler-rt/lib/builtins/modhi3.c
```diff
@@ -0,0 +1,22 @@
...
+//
+// This file implements __modhi3 for the compiler_rt library.
+//
+//===----------------------------------------------------------------------===//
+
+#include "int_lib.h"
+
+// Returns: a % b
+
+COMPILER_RT_ABI hi_int __modhi3(hi_int a, hi_int b) {
```
asl, line 17: Can we have it msp430 specific?

asl:
Can we have *hi builtins move to msp430 folder, please?

atrosinenko:
> Can we have *hi builtins move to msp430 folder, please?

Done. I have left their names unchanged (that is, traditional compiler-rt name aliased to __mspabi_*). As I understand, now they are "staging" but someday they may probably become part of 16-bit target support (and then be moved to common directory).

atrosinenko:
```
Failing Tests (11):
  Builtins-msp430-linux :: addtf3_test.c
  Builtins-msp430-linux :: comparedf2_test.c
  Builtins-msp430-linux :: compiler_rt_logbf_test.c
  Builtins-msp430-linux :: divdc3_test.c
  Builtins-msp430-linux :: divsc3_test.c
  Builtins-msp430-linux :: divtc3_test.c
  Builtins-msp430-linux :: muldc3_test.c
  Builtins-msp430-linux :: mulsc3_test.c
  Builtins-msp430-linux :: mulvdi3_test.c
  Builtins-msp430-linux :: subtf3_test.c
  Builtins-msp430-linux :: udivmoddi4_test.c

Testing Time: 9.08s
  Unsupported Tests  :  95
  Expected Passes    : 105
  Expected Failures  :   1
  Unexpected Failures:  11
```

asl:
I would concentrate on udivmoddi4_test.c

atrosinenko:
> I would concentrate on udivmoddi4_test.c

Chances are, it will "just work" when I ultimately manage to link it. The problem is enabling optimization refuses to "just work", so I'm investigating it.

```
/old_ssd/ast/soft/msp430gcc/lib/gcc/msp430-elf/8.3.1/../../../../bin/msp430-elf-ld: /old_ssd/ast/msp430-projects/llvm-project/build-compiler-rt-msp430/test/builtins/Unit/MSP430LinuxConfig/Output/udivmoddi4_test.c.tmp section `.data' will not fit in region `RAM'
/old_ssd/ast/soft/msp430gcc/lib/gcc/msp430-elf/8.3.1/../../../../bin/msp430-elf-ld: section __interrupt_vector_31 loaded at [000000000000fffe,000000000000ffff] overlaps section .data loaded at [00000000000049a8,00000000000a58d5]
/old_ssd/ast/soft/msp430gcc/lib/gcc/msp430-elf/8.3.1/../../../../bin/msp430-elf-ld: region `RAM' overflowed by 616676 bytes
clang-11: error: msp430-elf-ld command failed with exit code 1 (use -v to see invocation)
```

asl:
Why is it so large? What is so special in it as compared to other tests?

atrosinenko:
And after some research, the issue turned out to be `du_int tests[][4]` of size `0xa0e00` bytes :) So, in this particular case optimizer cannot help. Still I have to debug why enabling optimizations makes some tests fail.

atrosinenko:
Re-implemented 64-bit shift helpers. Turned out, Clang expects R8, R9, R10 registers to remain unchanged even for MSP430_BUILTIN calling convention notwithstanding the fact they contain lower 24 bits of the first argument. I'm not sure whether it is correct or not, but I have not found any explicit exceptions for these registers for modified calling convention for builtin functions.

In case R8-R10 should be considered caller-saved for MSP430_BUILTIN, then MSP430 LLVM codegen should be patched instead for compatibility with libgcc.

asl:
@atrosinenko You need to check which registers are caller-saved for this new CC. But usually argument registers are considered as scratch ;)

atrosinenko:
`addtf3_test` and `subtf3_test` cannot be compiled because
```
/path/to/subtf3_test.c:15:10: fatal error: 'fenv.h' file not found
#include <fenv.h>
         ^~~~~~~~
1 error generated.
```
I have two questions on this:
* do we have to have `fenv.h` file or can we just ignore this?
* we are not planning supporting `long double`s on MSP430, as I understand. But `addtf3.c`, `subtf3.c`, `divtf3.c`, `multf3.c` and `powitf2.c` are contained in the `GENERIC_SOURCES` variable in CMakeLists.txt while many other `tf` helpers are listed in `GENERIC_TF_SOURCES`. Are these 5 helpers special in some way or should they be moved to the latter list as well? These were introduced in commits 361519f4 and 61d5d38e except for `powitf3.c`.

atrosinenko:
There is another `long double`-related `divtc3` helper added by 47174a5a with commit message 
> Add a generic version of __divtc3.c for long double complex division.
> Mark the unit test as applying to all platforms.

asl:
Let us ignore FP tests for now, really.

atrosinenko:
After moving `long double`-related LibCalls to `GENERIC_TF_SOURCES` list and consistently marking appropriate helpers as msp430_builtin I got the following:
```
Failing Tests (6):
  Builtins-msp430-linux :: compiler_rt_logbf_test.c
  Builtins-msp430-linux :: divdc3_test.c
  Builtins-msp430-linux :: divsc3_test.c
  Builtins-msp430-linux :: extendhfsf2_test.c
  Builtins-msp430-linux :: mulsc3_test.c
  Builtins-msp430-linux :: udivmoddi4_test.c


Testing Time: 26.10s
  Unsupported Tests  : 100
  Expected Passes    : 105
  Expected Failures  :   1
  Unexpected Failures:   6
```
with `extendhfsf2_test` being depending on the optimization level.

To Do:
* mark `extendhfsf2` LibCall and corresponding unit test ARM-specific (?)
* check complex type ABI. According to the MSP430 EABI, part 2.5:
  > The _Complex types defined in the C99 standard are not supported for MSP430.
  > MSP430 does support the C++ complex class and its complex operators and std::complex<> template.
  
  Unfortunately, I'm not experienced with complex numbers in C yet
* Generalize msp430_builtin for all helpers from compiler-rt, if possible
* Split `udivmoddi4` into several tests on MSP430
  * I would not want creating inherently separate tests for regular platforms and low-memory ones, otherwise _technically_ they can evolve separately then. On one hand, they not expected to be modified anyway. On the other hand, if they will be modified in a diverging way - they are not covered by tests because they are tests itself :)
  * as I understand, the `tests` array contains 658944 bytes that is 10 times larger the entire address space, so there would be many parts
  * to not clutter the test directory with lots of `inc` files, I would rather prefer reading these testcases in binary form from `stdin`, if possible

asl:
> Split udivmoddi4 into several tests on MSP430
Can we disable this test for msp430 for now? :)

compiler-rt/lib/builtins/adddf3.c
```diff
@@ -13,7 +13,8 @@
 #define DOUBLE_PRECISION
 #include "fp_add_impl.inc"
 
-COMPILER_RT_ABI double __adddf3(double a, double b) { return __addXf3__(a, b); }
+COMPILER_RT_ABI double
```
asl, line 16: Whitespace-only change?

compiler-rt/lib/builtins/muldf3.c
```diff
@@ -14,7 +14,8 @@
 #define DOUBLE_PRECISION
 #include "fp_mul_impl.inc"
 
-COMPILER_RT_ABI fp_t __muldf3(fp_t a, fp_t b) { return __mulXf3__(a, b); }
+COMPILER_RT_ABI fp_t
```
asl, line 17: whitespace change?

compiler-rt/lib/builtins/udivmoddi4.c
```diff
@@ -82,7 +82,7 @@ COMPILER_RT_ABI du_int __udivmoddi4(du_int a, du_int b, du_int *rem) {
         r.s.high = n.s.high & (d.s.high - 1);
         *rem = r.all;
       }
-      return n.s.high >> __builtin_ctz(d.s.high);
```
asl, line 85: what is the problem with using __builtin?

compiler-rt/test/builtins/Unit/udivmoddi4_test.c
```diff
@@ -32,8 +32,13 @@ int test__udivmoddi4(du_int a, du_int b, du_int expected_q, du_int expected_r)
 
 char assumption_1[sizeof(di_int) == 2*sizeof(si_int)] = {0};
 
+// FIXME: Refactor this test so it can be linked on a platform
+// with 16-bit address space such as MSP430.
+// For now, simply use some arbitrary (not random) subset from the end of this
+// array.
 du_int tests[][4] =
 {
+#ifndef __MSP430__
```
asl, line 41: How about XFAIL: msp430?

llvm/lib/Target/MSP430/MSP430ISelLowering.cpp
```diff
@@ -448,6 +439,14 @@ static void AnalyzeVarArgs(CCState &State,
   State.AnalyzeFormalArguments(Ins, CC_MSP430_AssignStack);
 }
 
+static bool UseModifiedBuiltinCC(const CCState &State,
+                                 ArrayRef<unsigned> ArgsParts)
+{
```
asl, line 444: Put brace on preceding line

compiler-rt/lib/builtins/udivmoddi4.c
atrosinenko: The problem is `__builtin_ctz` has `int` aka `int16_t` argument.

compiler-rt/lib/builtins/muldf3.c
atrosinenko: Yes, because of 7dccccfa followed by 94a33d22. Will rollback that whitespace-only change.

atrosinenko:
When debugging divsc3_test, I have found some interesting things:
* isnan(float) is ultimately emitted via a call to __fpclassifyf that is linked from libc.a (and has another implementation of the same size inside libm.a). This happens when -fno-builtin is passed to clang. While it can be emitted implicitly (AFAIK, via __builtin_isnan), it resides in both libc and libm and not libgcc
* the real issue seems to be caused by `crealf` / `cimagf` functions

From `llvm-objdump -d` output:
```
00004cbc <cimagf>:
    4cbc: 0c 4e                         mov     r14, r12

00004cbe <.LVL1>:
    4cbe: 0d 4f                         mov     r15, r13
    4cc0: 30 41                         ret

00004cc2 <crealf>:
    4cc2: 30 41                         ret
```
So this simply looks like passing a `float _Complex` argument (that is 64 bit in total) in R12-R15 registers. Not too surprising at the first glance.

On the other hand, MSP430 EABI (SLAA534, June 2013) declares:
```
2.5 Complex Types

The _Complex types defined in the C99 standard are not supported for MSP430.
MSP430 does support the C++ complex class and its complex operators and std::complex<> template.

2.6 Structures and Unions
...
Structures having size 32 bits or less may reside in registers or register pairs while being passed to or returned from functions.
...
```

When checking with msp430-elf-gcc:
```
$ msp430-elf-gcc --version
msp430-elf-gcc (Mitto Systems Limited - msp430-gcc 8.3.1.25) 8.3.1
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
on a test derived from `crealf.c` and `cimagf.c` from the newlib:
```cpp
#include <complex.h>

typedef union {
        float complex z;
        float parts[2];
} float_complex;

float crealf(float complex z)
{
  float_complex w = { .z = z };
  return w.parts[0];
}

float cimagf(float complex z)
{
  float_complex w = { .z = z };
  return w.parts[1];
}

float crealf_explicit_1(float_complex w)
{
  return w.parts[0];
}

float crealf_explicit_2(float parts[2])
{
  return parts[0];
}
```
I get the wollowing:
```
$ msp430-elf-gcc -S test.c -o- -mmcu=msp430g2553 -Os
        .file   "crealf.c"
.text
        .balign 2
        .global crealf
        .type   crealf, @function
crealf:
; start of function
; attributes: 
; framesize_regs:     0
; framesize_locals:   0
; framesize_outgoing: 0
; framesize:          0
; elim ap -> fp       2
; elim fp -> sp       0
; saved regs:(none)
        ; start of prologue
        ; end of prologue
        ; start of epilogue
        RET
        .size   crealf, .-crealf
        .balign 2
        .global cimagf
        .type   cimagf, @function
cimagf:
; start of function
; attributes: 
; framesize_regs:     0
; framesize_locals:   0
; framesize_outgoing: 0
; framesize:          0
; elim ap -> fp       2
; elim fp -> sp       0
; saved regs:(none)
        ; start of prologue
        ; end of prologue
        MOV.W   R14, R12
        MOV.W   R15, R13
        ; start of epilogue
        RET
        .size   cimagf, .-cimagf
        .balign 2
        .global crealf_explicit_1
        .type   crealf_explicit_1, @function
crealf_explicit_1:
; start of function
; framesize_regs:     0
; framesize_locals:   0
; framesize_outgoing: 0
; framesize:          0
; elim ap -> fp       2
; elim fp -> sp       0
; saved regs:(none)
        ; start of prologue
        ; end of prologue
        MOV.W   R12, R13
        MOV.W   @R12, R12
        MOV.W   2(R13), R13
        ; start of epilogue
        RET
        .size   crealf_explicit_1, .-crealf_explicit_1
        .balign 2
        .global crealf_explicit_2
        .type   crealf_explicit_2, @function
crealf_explicit_2:
; start of function
; framesize_regs:     0
; framesize_locals:   0
; framesize_outgoing: 0
; framesize:          0
; elim ap -> fp       2
; elim fp -> sp       0
; saved regs:(none)
        ; start of prologue
        ; end of prologue
        MOV.W   R12, R13
        MOV.W   @R12, R12
        MOV.W   2(R13), R13
        ; start of epilogue
        RET
        .size   crealf_explicit_2, .-crealf_explicit_2
        .ident  "GCC: (Mitto Systems Limited - msp430-gcc 8.3.1.25) 8.3.1"
```
So, looks like this version of msp430-elf-gcc
1) **does** support `float _Complex` type and
2) **does** handle it in a special way

Not sure yet, maybe this EABI document is outdated now...

atrosinenko:
Squashed and rearranged commits a bit. Now, `git diff HEAD 7e3ee1a1e9f` shows nothing, so I expect final commit to represent the same tree state as before rebasing.

The first two commits could probably be sent beforehand if appropriate.

atrosinenko:
Current unit testing status:
```
Failing Tests (2):
  Builtins-msp430-linux :: compiler_rt_logbf_test.c
  Builtins-msp430-linux :: divdc3_test.c

Testing Time: 57.30s
  Unsupported Tests  : 101
  Expected Passes    : 108
  Expected Failures  :   1
  Unexpected Failures:   2
```

The `compiler_rt_logbf_test.c` test may be XFAIL due to oversimplified "reference" implementation from `libm` but still needs further investigation.

atrosinenko:
Turned out, UBSan can be useful even on MSP430 when using trap-on-error mode. `rep_clz` for single precision was always using `int` version of `__builtin_clz` that caused `__compiler_rt_logbf` to evaluate `0 - 127 - 0` (exp - exponentBias - shift) to 0.

atrosinenko:
The current version does not have any Unexpected Failures when testing with `-O0` to `-O3`, `-Os` and `-Oz` with my configuration. It still has to be checked against the original TI's libgcc (but some incompatibilities were already identified because the libc uses libgcc as well).

atrosinenko:
Looks like there are quite a lot helper functions with two 64-bit arguments that have a standard calling convention:

* `__addvdi3`, `__subvdi3` - they have `mov smth, r11` as a first instruction accessing `r11` register that obviously cannot be ignored for add/sub and the first argument in R8::R11
* `__mulvdi3`
* `__cmpdi2`, `__ucmpdi2` - the same as the first
* `__ledf2` and other these comparisons (but **not** weaker __mspabi_cmpd)

So, it seems quite the same amount of work is required to mark those helpers that use the modified calling convention and those not using while the first option seems not "inverted" at least.

atrosinenko:
Another issue found when trying to experiment with libgcc as a replacement for `clang_rt` were `__clzsi` and others returning `uint32_t`. As I understand, they should return plain `int` but on previously supported platforms it was 32-bit wide. Now, it can be 16-bit on MSP430, so I introduced a `native_int` typedef (this can signify that it is not just "unspecified width" more explicitly but I don't mind replacing it with `int`).

This change was quite massive and probably unfinished (for example, these helpers were mentioned somewhere in the sanitizer code) but should not probably break **existing** targets provided they all have 32-bit default `int` type. On the other hand, 16 bits are sufficient for all these function's returned values and it may be beneficial for 16-bit platforms by itself, apart from compatibility with libgcc.

PR #24, opened 2020-05-22
Fix __div[sdt]f3(minimal_normal_number, 2.0)

atrosinenko:
First, try aligning all three softfp division implementations.

My plan is to gradually make these three implementations textually as similar as possible. After that, extract them to `fp_div_impl.inc` file like it is already done for `__mul[sdt]f3` helpers.

When comparing them I have found some quite strange differences and I'm not sure whether it is intentional optimization or just a code that was copy-pasted and incorrectly updated. Unit tests for softfp division contain just two test cases each. Were they thoroughly debugged at all?..


atrosinenko:
Now, according to some fuzzing that compares __divdf3 results with ones obtained via native x86_64 CPU instructions, it correctly computes division result with the default rounding for float64. While the original question was about float128, current implementation is generic for all three sizes tuned with two defines: iterations count performed with half and full integer width.

atrosinenko:
_Just squashed into two commits. The next goal is to improve both commits by rewriting "in place"._

atrosinenko:
Just as a note, after ~130M of iterations, test_divtf3 failed with:
```
         a = 0x7FFF3463FAB38A79C8F703404AE51275
         b = 0xFFFFEBB15CC76E12857BCAFA1D8BC48C
         x = 0x7FFFB463FAB38A79C8F703404AE51275
  expected = 0xFFFFEBB15CC76E12857BCAFA1D8BC48C
```
As I understand, it is just some NaN that should be normally ignored by `compareResultLD`.

atrosinenko:
1. Squashed into three commits (add generally-applicable testcases, replace implementation, temporarily commit benchmarking code)
2. Replaced the middle commit into two:
   * first commit roughly modelling the original implementation in a unified manner
   * second commit adds the denormal result support

Now, the "NFC refactoring commit" passes basic unit tests but requires further proof revalidation and fuzzing. Still, it is useful for the purpose of performance debugging.

atrosinenko:
Another float128 issue, now not NaN-related:
```
         a = 0xFFFF630AA1E6ED1F319D498544DB38EA
         b = 0x7FFFDD22D65546B177698C9A7176766A
         x = 0xFFFFE30AA1E6ED1F319D498544DB38EA
  expected = 0x7FFFDD22D65546B177698C9A7176766A
```

atrosinenko:
Reviewed on Phabricator:
* https://reviews.llvm.org/D84932
* https://reviews.llvm.org/D85031
* https://reviews.llvm.org/D85032

Uploaded to mainline:
* [[compiler-rt][builtins] Add more test cases for __div[sdt]f3 LibCalls](https://github.com/llvm/llvm-project/commit/b9f49d13fd4400274ff24ecd847c710bf0c563c3)
* [[builtins] Unify the softfloat division implementation ](https://github.com/llvm/llvm-project/commit/0e90d8d4fed8c8cac70700acfdef6fc2ca2d482d)
* [[builtins] Make __div[sdt]f3 handle denormal results](https://github.com/llvm/llvm-project/commit/93eed63d2f31888d9d8eb0712287e44771289f6f)
* [[builtins] Fix divtf3_test.c](https://github.com/llvm/llvm-project/commit/553833958fdea48e41a11ee7e9c104c903deadf5)

Issue #26, opened 2020-06-08
Implement debug information emitting for MSP430


atrosinenko:
When debugging MSP430 code, it would be very useful to have at least the line number information. For example, it highly speeds up debugging where UBSan was triggered.

I have patched my working copy like this:
```diff
diff --git a/llvm/lib/Target/MSP430/MCTargetDesc/MSP430MCAsmInfo.cpp b/llvm/lib/Target/MSP430/MCTargetDesc/MSP430MCAsmInfo.cpp
index 6117567dcae..08928bdc785 100644
--- a/llvm/lib/Target/MSP430/MCTargetDesc/MSP430MCAsmInfo.cpp
+++ b/llvm/lib/Target/MSP430/MCTargetDesc/MSP430MCAsmInfo.cpp
@@ -24,4 +24,8 @@ MSP430MCAsmInfo::MSP430MCAsmInfo(const Triple &TT,
 
   AlignmentIsInBytes = false;
   UsesELFSectionDirectiveForBSS = true;
+
+  SupportsDebugInformation = true;
 }
diff --git a/llvm/lib/Target/MSP430/MSP430RegisterInfo.td b/llvm/lib/Target/MSP430/MSP430RegisterInfo.td
index 24d94c1739f..36f8970d582 100644
--- a/llvm/lib/Target/MSP430/MSP430RegisterInfo.td
+++ b/llvm/lib/Target/MSP430/MSP430RegisterInfo.td
@@ -19,11 +19,12 @@ class MSP430Reg<bits<4> num, string n, list<string> alt = []> : Register<n> {
 
 class MSP430RegWithSubregs<bits<4> num, string n, list<Register> subregs,
                            list<string> alt = []> 
-  : RegisterWithSubRegs<n, subregs> {
+  : RegisterWithSubRegs<n, subregs>, DwarfRegNum<[num]> {
   field bits<4> Num = num;
   let Namespace = "MSP430";
   let HWEncoding{3-0} = num;
   let AltNames = alt;
 }
 
 //===----------------------------------------------------------------------===//
```
And it seems to basically work. The DWARF3 register numbers are taken from MSP430 EABI document (subregister handling is still to-be-written). On the other hand, other targets such as X86 involve some non-trivial DWARF-related code in the TargetFrameLowering passes, for example.

If there is some well-known simple way to get at least basic debug info support in LLVM (possibly not too comprehensive but non-misleading), I would want to work on it.

asl:
[...]

PR #27, opened 2020-06-16
Emit DWARF for MSP430 target

atrosinenko:
Turned out, fixing #26 was not so trivial. While implementing several relocation types was quite easy, the real issue was the fact that LLVM emits the `.debug_info` with `addr_size` equal to `MCAsmInfo::CodePointerSize`. This even produces correct debug info (at least `msp430-elf-gdb` can use it to answer various questions about the program state).

The problems start when trying to test it. Looks like `llvm-dwarfdump` expects either DWARF32 with addr_size = 4 or DWARF64 with addr_size = 8, so it cannot print the contents of `.debug_info`. But it can dump debug info from object files produced by msp430-elf-gcc because it uses DWARF32 with addr_size = 4.

So, this PR tries to introduce another pointer size that is used for DWARF emitting. It may be worth to just change `CodePointerSize` from 2 to 4 for MSP430 (dropping the first of three commits and significantly simplifying the review) but at least `AsmPrinter::doFinalization` seems to use it exactly as a native pointer size.

Meanwhile, the same issue was with AVR target: I installed `avr-gcc` from system repository and it uses DWARF32 with addr_size = 4 as well. AVR backend for LLVM, on the other hand, cannot parse its own debug info. (What is pointer size on AVR32?)

On possible regressions:
* I need to recheck the update of `MSP430MCAsmInfo ` class according to LLVM guidelines (what compatibility is required and how to achieve it)
* this PR is expected to be NFC for targets with CodePointerSize >= 4
  * the only mainline targets that are not covered are MSP430 and AVR, both seems to function better with this patch
  * private target implementations can become broken, on the other hand (how is this handled?)

### DIFF
--- a/llvm/include/llvm/MC/MCAsmInfo.h
+++ b/llvm/include/llvm/MC/MCAsmInfo.h
@@ -424,6 +424,14 @@ public:
   /// Get the code pointer size in bytes.
   unsigned getCodePointerSize() const { return CodePointerSize; }
 
+  /// Get the pointer size in bytes to use when emitting DWARF.
+  /// On some targets where CodePointerSize == 2 (such as AVR and MSP430)
+  /// gcc emits .debug_info with addr_size = 0x04.
+  /// llvm-dwarfdump expects that as well to parse .debug_info.
+  unsigned getCodePointerSizeForDwarf() const {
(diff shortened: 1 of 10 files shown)


atrosinenko:
Rewritten in #28.

PR #31, opened 2020-08-01
[WIP] [MSP430] Fix passing structs as function arguments

atrosinenko:
This patch is expected to fix the ABI for passing `struct`s and `union`s as function arguments. For now, it just contains a skeleton of a unit test with comments on what should finally be implemented.

Interesting fact: this function

https://github.com/access-softek/llvm-project/blob/65fd651980a8ad965363807cc334c513e4c8ffe4/clang/lib/CodeGen/TargetInfo.cpp#L84-L88

is called with `ByRef == true`, as expected. But the second argument of `ABIArgInfo::getIndirect(...)` is `bool ByVal`:

https://github.com/access-softek/llvm-project/blob/65fd651980a8ad965363807cc334c513e4c8ffe4/clang/include/clang/CodeGen/CGFunctionInfo.h#L172-L182


atrosinenko:
Sent for public review: https://reviews.llvm.org/D85917

PR #62, opened 2023-12-04
[AArch64][PAC] Emit a fatal error when ptrauth target feature is missing

kovdan01:
Previously, in `expandPtrAuthPseudo` we expanded `MOVaddrPAC`, `LOADgotPAC` and `LOADauthptrgot` pseudo-instructions without taking presense of PAuth subtarget feature into account. In case of `LOADauthptrgot`, it resulted in undesired so-called `$auth_ptr$` stub and a corresponding AUTH relocation. In case `MOVaddrPAC` and `LOADgotPAC`, it resulted in pac-specific instructions (e.g. `paciza`) emitted which was only caught via assertion during machine instructions verification (and not caught at all when assertions are disabled).

This patch makes us emit a fatal error and fail fast in such cases.

### DIFF
--- a/llvm/test/CodeGen/AArch64/GlobalISel/ptrauth-lowering-err-debug.ll
+++ b/llvm/test/CodeGen/AArch64/GlobalISel/ptrauth-lowering-err-debug.ll
@@ -50,12 +50,11 @@ attributes #0 = { "ptrauth-calls" "target-cpu"="generic" }
 ; CHECK-NEXT: {{^$}}
 ; CHECK-NEXT: LLVM ERROR: pac instructions require ptrauth target feature
 
-define i8* @foo() #0 {
-  %tmp = bitcast { i8*, i32, i64, i64 }* @g_weak.ptrauth to i8*
-  ret i8* %tmp
+define ptr @foo() #0 {
+  ret ptr @g_weak.ptrauth
 }
 
 @g_weak = extern_weak global i32
-@g_weak.ptrauth = private constant { i8*, i32, i64, i64 } { i8* bitcast (i32* @g_weak to i8*), i32 0, i64 0, i64 0 }, section "llvm.ptrauth"
+@g_weak.ptrauth = private constant { ptr, i32, i64, i64 } { ptr @g_weak, i32 0, i64 0, i64 0 }, section "llvm.ptrauth"
 
 attributes #0 = { "ptrauth-calls" "target-cpu"="generic" }
--- a/llvm/test/CodeGen/AArch64/GlobalISel/ptrauth-lowering-err.ll
+++ b/llvm/test/CodeGen/AArch64/GlobalISel/ptrauth-lowering-err.ll
@@ -43,13 +43,12 @@ attributes #0 = { "ptrauth-calls" "target-cpu"="generic" }
 
 ; CHECK: LLVM ERROR: pac instructions require ptrauth target feature
 
-define i8* @foo() #0 {
-  %tmp = bitcast { i8*, i32, i64, i64 }* @g_weak.ptrauth to i8*
-  ret i8* %tmp
+define ptr @foo() #0 {
+  ret ptr @g_weak.ptrauth
 }
 
 @g_weak = extern_weak global i32
-@g_weak.ptrauth = private constant { i8*, i32, i64, i64 } { i8* bitcast (i32* @g_weak to i8*), i32 0, i64 0, i64 0 }, section "llvm.ptrauth"
+@g_weak.ptrauth = private constant { ptr, i32, i64, i64 } { ptr @g_weak, i32 0, i64 0, i64 0 }, section "llvm.ptrauth"
 
 attributes #0 = { "ptrauth-calls" "target-cpu"="generic" }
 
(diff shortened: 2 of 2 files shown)


asl:
@atrosinenko please review

llvm/test/CodeGen/AArch64/GlobalISel/ptrauth-lowering-err-debug.ll
```diff
@@ -0,0 +1,58 @@
+; REQUIRES: asserts
+; RUN: split-file %s %t && cd %t
+
+;--- MOVaddrPAC.ll
+
+; RUN: not --crash llc -debug -mtriple aarch64-elf MOVaddrPAC.ll 2>&1 | \
```
atrosinenko, line 6: `-debug` may be replaced with `-debug-only=aarch64-expand-hardened-pseudos` for less verbose debug output (not sure which is more canonical in tests, though).

atrosinenko (commented):
I wonder if other code paths have to be guarded as well.

llvm/test/CodeGen/AArch64/GlobalISel/ptrauth-lowering-err-debug.ll
asl: +1

kovdan01: Thanks, fixed in 52afe428f699b1eddcd6b649058d1e21c2f21b1b

llvm/test/CodeGen/AArch64/GlobalISel/ptrauth-lowering-err-debug.ll
```diff
@@ -0,0 +1,61 @@
...
+; CHECK-NEXT: {{^$}}
+; CHECK-NEXT: LLVM ERROR: pac instructions require ptrauth target feature
+
+define i8* @foo() #0 {
+  %tmp = bitcast { i8*, i32, i64, i64 }* @g_weak.ptrauth to i8*
+  ret i8* %tmp
+}
+
+@g_weak = extern_weak global i32
+@g_weak.ptrauth = private constant { i8*, i32, i64, i64 } { i8* bitcast (i32* @g_weak to i8*), i32 0, i64 0, i64 0 }, section "llvm.ptrauth"
```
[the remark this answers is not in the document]
kovdan01: Fixed in 8e30a83d8afda9d3ac963717f1047d14e54247c1

llvm/test/CodeGen/AArch64/GlobalISel/ptrauth-lowering-err.ll
```diff
@@ -0,0 +1,86 @@
...
+
+; CHECK: LLVM ERROR: pac instructions require ptrauth target feature
+
+define i8* @foo() #0 {
+  %tmp = bitcast { i8*, i32, i64, i64 }* @g_weak.ptrauth to i8*
+  ret i8* %tmp
+}
+
+@g_weak = extern_weak global i32
+@g_weak.ptrauth = private constant { i8*, i32, i64, i64 } { i8* bitcast (i32* @g_weak to i8*), i32 0, i64 0, i64 0 }, section "llvm.ptrauth"
```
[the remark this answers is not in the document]
kovdan01: Fixed in 8e30a83d8afda9d3ac963717f1047d14e54247c1

kovdan01:
> Is `AArch64ExpandHardenedPseudos::expandAuthLoad` method expected to contain a similar check?

@atrosinenko In bffc609c13d5f077e29a5bb1ca1cb4f7022520dc I've added checks in `AArch64DAGToDAGISel::tryAuthLoad` and before `AArch64InstructionSelector::selectAuthLoad` - those are the places where `LDRA*` pseudos are emitted. So, we should only go to `AArch64ExpandHardenedPseudos::expandAuthLoad` after these checks - inserted a corresponding assertion there.

atrosinenko (commented):
Sorry for too many iterations, I just spotted the tests are in the `llvm/lib/Target/AArch64/GISel/` directory. Maybe these could be moved to `llvm/lib/Target/AArch64` and updated to test both DAGISel and GlobalISel like it is done in [ptrauth-intrinsic-blend.ll](https://github.com/llvm/llvm-project/blob/main/llvm/test/CodeGen/AArch64/ptrauth-intrinsic-blend.ll). This should make `TODO` unneeded, I guess.

kovdan01:
@atrosinenko Thanks, testing both with `-global-isel=0` and `-global-isel=1` is definitely needed. It revealed several new cases with failures.

I don't feel the approach initially taken suites best now - trying to guard every place where we potentially emit auth-specific instructions makes things harder to maintain, understand, and there are way too many points responsible for checking such a simple things.

I suppose the following approach could be better.
- For each function, if PAuth subtarget feature is not enabled, see if there are "ptrauth-*" attributes ("ptrauth-calls", "ptrauth-returns", etc). We can implement a separate pass for such a check.
- If PAuth subtarget feature is not enabled, check if there are any objects in "llvm.ptrauth" section.
- Check llvm ptrauth intrinsics separately.

Please let me know your thoughts on this. Thanks!

kovdan01:
As discussed with @asl, we should move the checks to frontend. See https://github.com/access-softek/llvm-project/issues/65.

PR #64, opened 2023-12-11
[WIP][PAC] Emit IR debug info metadata nodes for implicitly signed pointers

kovdan01:
Emit debug metadata nodes with `DW_TAG_LLVM_ptrauth_type` tag containing signing scheme info for the following types of implicitly (without `__ptrauth` specifier) signed pointers:

- virtual table pointers;
- free function pointers;
- member function pointers.


asl:
Tagging @ahmedbougacha

kovdan01:
In any case, we need to get info about ABI signing schema used by debuggee in order to pass desired flags to compiler in context of user expression evaluation. Given that, if we know the ABI, there is no reason in emitting ptrauth dwarf data for implicitly signed pointers - the compiler will sign them properly when we pass signing flags used by ABI to the compiler. So, in Dwarf we only need to have info about explicitly signed pointers. Closing this as not needed anymore.

PR #66, opened 2023-12-21
Error if PAuth language features are used on unsupported CPU

atrosinenko:
Check for several -fptrauth-* language features that require CPU with FEAT_PAuth support. If such unsupported feature is requested, make Clang frontend print error and exit early.


atrosinenko:
Looks like `HasPAuth` and a few other flags should be conditionally set in `AArch64TargetInfo::setArchFeatures` function, but they don't: https://github.com/access-softek/llvm-project/commit/9e3a02b1661a17f8f6815eec5fc783fbd92ff00f Strictly speaking, these flags do not directly correspond to FEAT_* identifiers (example: `HasCCPP`), so this should be checked a bit more...

clang/lib/Basic/PointerAuthOptions.cpp
```diff
@@ -0,0 +1,34 @@
...
+  Result.push_back(CXXTypeInfoVTablePointer);
+  Result.push_back(CXXVTTVTablePointers);
+  Result.push_back(CXXVirtualFunctionPointers);
+  Result.push_back(CXXVirtualVariadicFunctionPointers);
+  Result.push_back(CXXMemberFunctionPointers);
+  Result.push_back(ObjCIsaPointers);
+  Result.push_back(ObjCSuperPointers);
+  Result.push_back(InitFiniPointers);
+  assert(sizeof(Result[0]) * Result.size() + 4 == sizeof(PointerAuthOptions));
+  return Result;
```
atrosinenko, line 33: This definitely needs refactoring, but I didn't come up with a better way to enumerate all these `PointerAuthSchema`s yet.

atrosinenko:
As far as I can see, it is Clang **driver** that sets the per-triple default CPU and features, but `clang -cc1` enables several pauth-related features based on `arm64e-*` triple. Thus, I have a number of options:
* enable pauth in driver instead, fix the affected cc1 tests by adding the new options enabling pointer authentication in `clang -cc1` tests
* explicitly pass `-target-feature +pauth` to `clang -cc1`, where needed, to make existing tests pass
* emit error messages inside the driver, not frontend (the driver probably lacks available CPU features in parsed form)

I plan implementing the second variant (add target-feature / target-cpu to existing tests). This seems safer than the first option (less probability of accidentally disabling pointer authentication).

asl:
@atrosinenko I think everything should be done in driver. Tests are likely to be fixed.

atrosinenko:
> As far as I can see, it is Clang driver that sets the per-triple default CPU and features, but clang -cc1 enables several pauth-related features based on arm64e-* triple.

Sorry, this seems to be incorrect: both choosing the default CPU/features and `-fptrauth-...` options takes place in the driver, as expected. The only change in ~70 tests is to insert `-target-feature +pauth` options in invocations like this:

```
// RUN: %clang_cc1 -target arm64e-apple-ios -fptrauth-calls ... -emit-llvm ...
```

(i.e. those enabling ptrauth and not involving actual code generation either in assembler or object file form). Several tests still have to be fixed.

Maybe it is better to check for unsupported options in the driver instead. This should eliminate `clang::PointerAuthOptions::getAllUsedSchemas()` but may require more complicated checking for presence of `+pauth` feature. I will check this tomorrow.

atrosinenko:
A few notes:
* In this patch, `-fptrauth-intrinsics` is considered as requiring `+pauth` feature. Some `ptrauth_*` intrinsics can be implemented by only using HINT-encoded opcodes (at least, signing and authentication using IA and IB keys). I guess, this can be useful in libraries such as musl. When upstreaming, it may be better to replace the check for `Opts.PointerAuthIntrinsics` with per-intrinsic checks at some other places.
* The error message is usually printed twice (when `AArch64TargetInfo::adjust` is called from `CompilerInstance::createTarget` and from `CompilerInstance::createPreprocessor`). Hopefully, this is not a big issue at now.
* This PR contains a commit that sets `HasPAuth` flag in `AArch64TargetInfo::setArchFeatures` if v8.3+ or v9.0+ extensions are enabled. It is worth sending a more elaborate patch upstream (there are a few other seemingly forgotten flags)

clang/include/clang/Basic/DiagnosticDriverKinds.td
```diff
@@ -532,6 +532,9 @@ def err_stack_tagging_requires_hardware_feature : Error<
   "'-fsanitize=memtag-stack' requires hardware support (+memtag). For Armv8 or "
   "Armv9, try compiling with -march=armv8a+memtag or -march=armv9a+memtag">;
 
+def err_pauth_cpu_feature_missing : Error<
+  "A feature is requested that needs CPU with Pointer Authentication support">;
```
kovdan01, line 536: Nit: according to llvm coding standards, error and warning messages should start with a lower-case letter https://llvm.org/docs/CodingStandards.html#error-and-warning-messages.

The message text looks OK, but I'll think about it a bit more - I believe it can be re-phrased in a better way (already good enough now though).

clang/test/CodeGen/ptrauth-cpu-feature.c
```diff
@@ -0,0 +1,59 @@
...
+// Test that features requiring FEAT_PAuth fail early if the requirement is not met:
+//
+// RUN: not %clang %s -S -o - -target aarch64-linux-gnu -mcpu=cortex-a72 -mbranch-protection=pauthabi 2>&1 | FileCheck %s --check-prefix=FAIL
+// RUN: not %clang %s -S -o - -target aarch64-linux-gnu -mcpu=cortex-a72 -fptrauth-calls              2>&1 | FileCheck %s --check-prefix=FAIL
+// RUN: not %clang %s -S -o - -target aarch64-linux-gnu -mcpu=cortex-a72 -fptrauth-returns            2>&1 | FileCheck %s --check-prefix=FAIL
+// RUN: not %clang %s -S -o - -target aarch64-linux-gnu -mcpu=cortex-a72 -fptrauth-objc-isa           2>&1 | FileCheck %s --check-prefix=FAIL
+//
+// RUN:     %clang %s -S -o - -target aarch64-linux-gnu -mcpu=apple-a12  -mbranch-protection=pauthabi 2>&1 | FileCheck %s --check-prefix=PAUTH
+
+// Test that the following options are still gated on -fptrauth-calls.
```
kovdan01, line 12: Don't we want to emit a warning or smth in these cases? A user probably meant to enable some PAuth-related feature if one of these options is passed. If we just silently do not actually enable the feature unless `-fptrauth-calls` is here, it might be surprising for the user.

clang/lib/Basic/Targets/AArch64.cpp
```diff
@@ -197,6 +200,29 @@ AArch64TargetInfo::AArch64TargetInfo(const llvm::Triple &Triple,
         Opts.EABIVersion == llvm::EABI::GNU ? "\01_mcount" : "mcount";
 }
 
+void AArch64TargetInfo::validatePAuthOptions(DiagnosticsEngine &Diags,
+                                             const LangOptions &Opts) const {
+  if (Opts.SoftPointerAuth)
+    return;
+
+  bool NeedsFeatPAuth = false;
+  // FIXME: Some ptrauth_* intrinsics can be implemented without FEAT_PAuth.
```
kovdan01, line 209: Is this still true even with `Opts.SoftPointerAuth == false`? Isn't the if statement above enough?

clang/test/CodeGen/ptrauth-cpu-feature.c
```diff
@@ -0,0 +1,59 @@
+// REQUIRES: aarch64-registered-target
```
kovdan01, line 1: A couple of test cases with `-NOT` suffix for FileCheck directives can be added. Particularly, ensure that with major == 8, minor >= 3 or major >= 9 the feature is enabled and the error message is not produced.

clang/test/CodeGen/ptrauth-cpu-feature.c
```diff
@@ -0,0 +1,59 @@
...
+
+// Test that features requiring FEAT_PAuth fail early if the requirement is not met:
+//
+// RUN: not %clang %s -S -o - -target aarch64-linux-gnu -mcpu=cortex-a72 -mbranch-protection=pauthabi 2>&1 | FileCheck %s --check-prefix=FAIL
```
kovdan01, line 5: It looks like a more generic target triple should be enough (aarch64-elf or even just aarch64).

kovdan01 (changes requested):

clang/lib/Basic/Targets/AArch64.cpp
atrosinenko: This comment is about operations like sign and auth using IA or IB keys - these can be performed with only HINT-encoded instructions (but this is not the case for DA and DB keys). I guess these may be useful in some run-time libraries. (If they are definitely not useful, it may be worth commenting llvm/llvm-project#74729 and llvm/llvm-project#72502 as well)

atrosinenko:
llvm/llvm-project#78027 is the mainline PR that should replace 24048e11b427958315687a330e2ea359e04de9be commit in this PR.

clang/include/clang/Basic/DiagnosticDriverKinds.td
atrosinenko: Thanks! Updated the message.

clang/lib/Basic/Targets/AArch64.cpp
atrosinenko: Fixed

clang/test/CodeGen/ptrauth-cpu-feature.c
atrosinenko: Agree, added a warning

clang/test/CodeGen/ptrauth-cpu-feature.c
atrosinenko: Fixed

atrosinenko:
Addressed the comments, except for using a more generic triple in the test. A few weeks ago I saw a Discourse thread on "generic vs. precise triples" - obviously "the more triples tested the better", but I cannot remember whether that was an argument for using generic triples or for picking an arbitrary triple that is the most familiar for the particular developer...  I guess that it is better to have as precise triple as possible (even though arbitrarily chosen at commit time) for better reproducibility. Links to coding standards and Discourse threads are welcome.

Note: after all review comments would be ultimately addressed, I would like to *manually* squash and rebase, so that the commit that sets `HasPAuth` flag remains a separate commit to be rolled back and then replaced by mainline one (see https://github.com/llvm/llvm-project/pull/78027).

clang/lib/Basic/Targets/AArch64.cpp
```diff
@@ -202,19 +202,40 @@ AArch64TargetInfo::AArch64TargetInfo(const llvm::Triple &Triple,
...
+  //        HINT-encoded instructions, thus not requiring FEAT_PAUTH.
   //        At now, checking conservatively.
   NeedsFeatPAuth |= Opts.PointerAuthIntrinsics;
   NeedsFeatPAuth |= Opts.PointerAuthCalls;
   NeedsFeatPAuth |= Opts.PointerAuthReturns;
   NeedsFeatPAuth |= Opts.getPointerAuthObjcIsaAuthentication() !=
                     PointerAuthenticationMode::None;
 
-  if (NeedsFeatPAuth && !HasPAuth)
+  // FIXME The particular command line option is not printed in error message,
```
kovdan01, line 235: We already have a warning `warn_pauth_option_ignored` where we say that either `-fptrauth-calls` or `-mbranch-protection=pauthabi` should be enabled. Why not printing a similar error message mentioning both particular option (say, `-fptrauth-returns`) and the `-mbranch-protection=pauthabi` which includes the `-fptrauth-*` options? I don't feel that "misleadingly" in your comment fits here - one might want to enable only a particular feature and not the whole `-mbranch-protection` stuff, so, mentioning both in error message (like in `warn_pauth_option_ignored`) looks reasonable as for me

clang/include/clang/Basic/DiagnosticDriverKinds.td
```diff
@@ -533,7 +533,11 @@ def err_stack_tagging_requires_hardware_feature : Error<
   "Armv9, try compiling with -march=armv8a+memtag or -march=armv9a+memtag">;
 
 def err_pauth_cpu_feature_missing : Error<
-  "A feature is requested that needs CPU with Pointer Authentication support">;
+  "neither FEAT_PAUTH nor -fptrauth-soft is enabled. Most of PAuth features are unavailable">;
```
kovdan01, line 536: I already left a comment in clang/lib/Basic/Targets/AArch64.cpp and fixing that might make the current comment obsolete, but leaving it here anyway. It looks like that a **feature** is *enabled*, while an **option** is *specified* or *passed*. So, the following might be better: "neither FEAT_PAUTH is enabled nor -fptrauth-soft is passed..."

clang/test/CodeGen/ptrauth-cpu-feature.c
kovdan01: Some time ago I had a revision where MaskRay asked me to use a generic triple: see the comment https://reviews.llvm.org/D156882?id=546424#inline-1518496. I agree that specific triples are better for reproduceability, but is it OK in our particular context that running commands with different aarch64 triples gives different results? If results in all the cases must be equivalent, having a generic triple looks clearer IMHO.

clang/lib/Basic/Targets/AArch64.cpp
atrosinenko: Updated the message, thanks. Additionally, message duplication was fixed: now ignored/unsupported options are unset "on first iteration", so they do not trigger diagnostics anymore. Surprisingly, this slightly changed the behavior (due to `__has_feature(...)` preprocessor macro), but hopefully this is OK.

clang/include/clang/Basic/DiagnosticDriverKinds.td
atrosinenko: Fixed

atrosinenko:
Once again, addressed everything except target triples :) Thanks for the link - I will update the test after a bit more research.

atrosinenko:
Changed the cases that do not rely on features not being implicitly enabled to `-target aarch64` (assuming no features are ever implicitly *disabled* due to target-specific defaults).

kovdan01 (approved):
LGTM

atrosinenko:
Dropped the commit that sets `HasPAuth` is v8.3+ (replaced by #74), squashed and rebased.

PR #74, opened 2024-02-05
[AArch64] Make +pauth enabled in Armv8.3-a by default (#78027)

atrosinenko:
NB: This is just a cherry-pick of the
    a52eea66795018550e95c4b060165a7250899298 mainline commit.
    Due to the structure of the involved files, cherry-picking required
    fixing lots of trivial merge conflicts here and there.
    If you are rebasing elf-pauth branch on top of the mainline `main`
    branch after the original commit, most probably you can just skip
    this cherry-picked one.

Add AEK_PAUTH to ARMV8_3A in TargetParser and let it propagate to ARMV8R, as it aligns with GCC defaults.

After adding AEK_PAUTH, several tests from TargetParserTest.cpp crashed when trying to format an error message, thus update a format string in AssertSameExtensionFlags to account for bitmask being pre-formatted as std::string.

The CHECK-PAUTH* lines in aarch64-target-features.c are updated to account for the fact that FEAT_PAUTH support and pac-ret can be enabled independently and all four combinations are possible.

### DIFF
--- a/clang/lib/Basic/Targets/AArch64.cpp
+++ b/clang/lib/Basic/Targets/AArch64.cpp
@@ -257,7 +257,6 @@ void AArch64TargetInfo::getTargetDefinesARMV83A(const LangOptions &Opts,
                                                 MacroBuilder &Builder) const {
   Builder.defineMacro("__ARM_FEATURE_COMPLEX", "1");
   Builder.defineMacro("__ARM_FEATURE_JCVT", "1");
-  Builder.defineMacro("__ARM_FEATURE_PAUTH", "1");
   // Also include the Armv8.2 defines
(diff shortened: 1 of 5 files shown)


kovdan01 (approved):
I've not dove deeply into ARM documentation since it was already done by reviewers in original mainline PR and I assume that enough (please let me know if you want me to perform the review "from scratch" just like guys in mainline did).

In terms of cherry-picking and fixing merge conflicts and any stuff which differs from the mainline - LGTM.

atrosinenko:
@kovdan01

> I've not dove deeply into ARM documentation since it was already done by reviewers in original mainline PR and I assume that enough

This is exactly as intended, thank you.

PR #77, opened 2024-04-22
Dkovalev/pauth signed got mainline

kovdan01:


lld/ELF/SyntheticSections.cpp
```diff
@@ -718,6 +722,21 @@ void GotSection::writeTo(uint8_t *buf) {
     return;
   target->writeGotHeader(buf);
   target->relocateAlloc(*this, buf);
+  for (const AuthEntryInfo &authEntry : authEntries) {
+    // https://github.com/ARM-software/abi-aa/blob/main/pauthabielf64/pauthabielf64.rst#default-signing-schema
+    //   Signed GOT entries use the IA key for symbols of type STT_FUNC and the
+    //   DA key for all other symbol types, with the address of the GOT entry as
+    //   the modifier. The static linker must encode the signing schema into the
```
asl, line 729:
```suggestion
    //   the discriminator. The static linker must encode the signing schema into the
```

kovdan01: In the ABI, they use the modifier term in this particular paragraph. Do we want to stick with discriminator everywhere for consistency or maybe we want to use terms in the same way as in the ABI? It looks like that they use discriminator for the 16-bit value and modifier for the value with address optionally blended:

> discriminator is a 16-bit unsigned integer that after an optional blending (address diversity) forms the modifier for the sign and authenticate instructions.

asl: You can probably use the term "address diversity"?

kovdan01: Possible variant of paraphrasing this - please let me know if it's OK. Before:

> Signed GOT entries use the IA key for symbols of type STT_FUNC and the DA key for all other symbol types, with the address of the GOT entry as the modifier.

After:

> Signed GOT entries use the IA key for symbols of type STT_FUNC and the DA key for all other symbol types, with address diversity enabled and no additional 16-bit discriminator.

I'm not sure if it's any better, but, if the "modifier" term confuses code readers, I have no issues with avoiding it (but it's anyway already widely used in the spec so the person should be familiar with it).

asl: Look better to me.

PR #112, opened 2026-04-25
[mlir][dxsa] Add dcl_global_flags instruction

tagolog:
Global shader flags (refactoringAllowed, enableDoublePrecision, etc.) represented as named enum cases, making the MLIR output human-readable:

  dxsa.dcl_global_flags <refactoringAllowed|enableDoublePrecision>


asavonic (approved):
Excellent, thank you!

mlir/test/Target/DXSA/dcl_global_flags_none.mlir
```diff
@@ -0,0 +1,3 @@
+// RUN: not mlir-translate --import-dxsa-bin %S/inputs/dcl_global_flags_none.bin 2>&1 | FileCheck %s
```
asl, line 1: As we discussed, this is not a proper way of checking diagnostics. https://mlir.llvm.org/getting_started/TestingGuide/#diagnostic-tests Also, internal IR consistency checks should not be used for file format verification

PR #113, opened 2026-04-27
[mlir][dxsa] Add BinaryWriter to translate from MLIR to DXSA

asavonic:
BinaryWriter translates from an MLIR module in DXSA dialect into a DXSA binary. It is a reverse of what BinaryParser does.

Current implementation only supports standard instructions, and needs to be extended to support custom instructions.

Instruction table is moved into a separate file (InstrInfo.def), so it can be shared between Parser/Writer, which build different data structures from it. Parser goes from opcodes to mnemonics, and Writer is reversed.

Tests are extended to run MLIR in roundtrip to verify both the Parser and Writer. We also compare binary output with input to make sure that we do not lose any data during translation.


mlir/test/Target/DXSA/empty.mlir
```diff
@@ -1,4 +1,8 @@
 // RUN: mlir-translate --import-dxsa-bin | FileCheck %s
+// RUN: mlir-translate --export-dxsa-bin %s -o - | mlir-translate --import-dxsa-bin - | FileCheck %s
```
asl, line 2: You can probably introduce `--verify-roundtrip`, similar to what `mlir-op` does for normal MLIR

mlir/lib/Target/DXSA/BinaryWriter.cpp
```diff
@@ -17,10 +17,359 @@
...
+public:
+  Writer(raw_ostream &output) : output(output, endianness::little) {
+    initOpcodeMap(opcodeMap);
+  }
+
+  LogicalResult emitModule(ModuleOp source) {
+    Region &region = source.getRegion();
+    if (!region.hasOneBlock()) {
+      emitError(region.getLoc(), "region should contain only one block");
+      return failure();
```
asl: `emitError` produces `InFlightDiagnostics` that is implicitly convertible to `LogicalResult`. This could be simplified down to:
```c++
  return emitError(region.getLoc() , "region should contain only one block");
```
Ditto in other places.

mlir/lib/Target/DXSA/BinaryWriter.cpp
```diff
@@ -17,10 +17,359 @@
...
+    }
+
+    for (auto &op : region.front()) {
+      if (auto inst = dyn_cast<dxsa::Instruction>(op)) {
+        if (failed(emitInstruction(inst))) {
```
asl, line 63: When doing something like this, please ensure that all paths of `emitInstruction()` produce meaningful diagnostics. Otherwise, you'll end with failed run w/o proper understanding what was wrong.

mlir/lib/Target/DXSA/BinaryWriter.cpp
```diff
@@ -17,10 +17,359 @@
...
+    buffer.push_back(token);
+
+    for (Value value : inst.getOperands()) {
+      Operation *op = value.getDefiningOp();
+      if (!op) {
+        emitError(value.getLoc(), "undefined operand");
+        return failure();
+      }
+
+      if (auto operand = dyn_cast<dxsa::Operand>(*op)) {
```
asl: You may probably want to use `TypeSwitch` here.

mlir/lib/Target/DXSA/BinaryWriter.cpp
```diff
@@ -17,10 +17,359 @@
...
+    uint32_t dim = 0;
+    for (Value value : op.getOperands()) {
+      Operation *index = value.getDefiningOp();
+      if (!index) {
+        emitError(value.getLoc(), "index must be defined");
+        return failure();
+      }
+
+      FailureOr<uint32_t> repr = getIndexRepresentation(index);
+      if (failed(repr)) {
```
asl: Is it really a "normal" error"? How can we have value out of the thin air?

mlir/lib/Target/DXSA/BinaryWriter.cpp
```diff
@@ -17,10 +17,359 @@
...
+
+    // Indices follow the operand token.
+    for (Value value : op.getOperands()) {
+      Operation *index = value.getDefiningOp();
+      if (!index) {
+        emitError(value.getLoc(), "index must be defined");
+        return failure();
+      }
+
+      if (auto indexImm = dyn_cast<dxsa::IndexImm>(*index)) {
```
asl: See above wrt `TypeSwitch`

mlir/lib/Target/DXSA/BinaryWriter.cpp
```diff
@@ -17,10 +17,359 @@
...
+    }
+
+    return success();
+  }
+
+  // Emit an immediate operand. Unlike register operands, immediate
+  // operands do not have indices. They are encoded as an operand
+  // followed by N immediate values for each component.
+  LogicalResult emitOperandImm(dxsa::OperandImm op) {
+    auto attr = dyn_cast<DenseIntElementsAttr>(op.getImm());
```
asl: do we have `verify` on `dxsa::OperandImm` that ensures proper type here? If not, then it should be added and / or type constraint added to the instruction definition.

mlir/lib/Target/DXSA/BinaryWriter.cpp
```diff
@@ -17,10 +17,359 @@
...
+    } else {
+      emitError(op.getLoc(), "invalid immediate operand type");
+      return failure();
+    }
+
+    // Split immediates into tokens. 32 bit immediate values are
+    // encoded as is, and 64 bit immediates are split into high and
+    // low 32 bit parts.
+    SmallVector<uint32_t, 4> values;
+    for (APInt v : attr) {
```
asl: you might probably want to grab things by const reference in order not to copy every `APInt` here.

mlir/lib/Target/DXSA/BinaryWriter.cpp
```diff
@@ -17,10 +17,359 @@
...
+      token |=
+          ENCODE_D3D10_SB_OPERAND_NUM_COMPONENTS(D3D10_SB_OPERAND_4_COMPONENT);
+    } else {
+      emitError(op.getLoc(),
+                "immediate operand should be either 1- or 4- component");
+      return failure();
+    }
+
+    buffer.push_back(token);
+    for (uint32_t v : values) {
```
asl: `llvm::append_range(buffer, values)`.

mlir/test/Target/DXSA/empty.mlir
asavonic: Good idea. I'll see how to add an option like this.

mlir/lib/Target/DXSA/BinaryWriter.cpp
asavonic: Great! I missed that.

mlir/lib/Target/DXSA/BinaryWriter.cpp
asavonic: Sure, that is expected. These patterns only propagate an error up.

mlir/lib/Target/DXSA/BinaryWriter.cpp
asavonic: Added `TypeSwitch` here and for indices as well.

mlir/lib/Target/DXSA/BinaryWriter.cpp
asavonic: It is not a normal error. I suppose it can only happen if a value is a block argument, but we do not have control flow for this dialect. Do you suggest to make it an assertion?

mlir/lib/Target/DXSA/BinaryWriter.cpp
asavonic: There is `AnyAttrOf<[I32ElementsAttr, I64ElementsAttr]>` constraint, but `getImm` returns `mlir::Attribute`, so we need to cast it here. Perhaps we can have an assertion instead, but I'm not sure if anything other than `DenseIntElementsAttr` is possible.

mlir/lib/Target/DXSA/BinaryWriter.cpp
asavonic: Done.

mlir/lib/Target/DXSA/BinaryWriter.cpp
asavonic: Done.

asavonic:
Thanks a lot Anton! Let me know if I can improve anything else.
We have some of these problems in `BinaryParser` as well. I'll fix them separately.

mlir/lib/Target/DXSA/BinaryWriter.cpp
asl: This should be enforced by IR then. I mean, if we're having graph regions of SSA dominance.

mlir/lib/Target/DXSA/BinaryWriter.cpp
asl: here is how the things are defined:
```tablegen
class SignlessIntElementsAttr<int width> : IntElementsAttrBase<
  CPred<"::llvm::cast<::mlir::DenseIntElementsAttr>($_self).getType()."
        "getElementType().isSignlessInteger(" # width # ")">,
  width # "-bit signless integer elements attribute"> {

  // Note that this is only constructing scalar elements attribute.
  let constBuilderCall = "::llvm::cast<::mlir::DenseIntElementsAttr>("
  "::mlir::DenseElementsAttr::get("
    "::mlir::RankedTensorType::get({}, $_builder.getIntegerType(" # width # ")), "
    "::llvm::ArrayRef($0)))";
}

def I32ElementsAttr : SignlessIntElementsAttr<32>;
def I64ElementsAttr : SignlessIntElementsAttr<64>;
```

So it is always `DenseIntElementsAttr` and this is enforced by verifier. You should just do direct `cast` here, no check / error is needed.

mlir/lib/Target/DXSA/BinaryWriter.cpp
```diff
@@ -17,10 +18,328 @@
...
+
+    uint32_t token = 0;
+
+    Type elementType = attr.getType().getElementType();
+    if (elementType.isInteger(32)) {
+      token |= ENCODE_D3D10_SB_OPERAND_TYPE(D3D10_SB_OPERAND_TYPE_IMMEDIATE32);
+    } else if (elementType.isInteger(64)) {
+      token |= ENCODE_D3D10_SB_OPERAND_TYPE(D3D10_SB_OPERAND_TYPE_IMMEDIATE64);
+    } else {
+      return emitError(op.getLoc(), "invalid immediate operand type");
```
asl: see above. You can probably also `TypeSwitch` here with unreachable on default case.

mlir/lib/Target/DXSA/BinaryWriter.cpp
```diff
@@ -17,10 +18,328 @@
...
+    buffer.push_back(token);
+    llvm::append_range(buffer, values);
+
+    return success();
+  }
+
+  // Emit an immediate index. Its type is encoded into the operand, so
+  // here we only emit the value as tokens.
+  LogicalResult emitIndexImm(dxsa::IndexImm op) {
+    auto attr = dyn_cast<IntegerAttr>(op.getImm());
```
asl: See above on `cast` thing. We can always assume that IR is valid, IR verifier enforces this. So, you'd assert instead of error out.

mlir/lib/Target/DXSA/BinaryWriter.cpp
```diff
@@ -17,10 +18,328 @@
...
+  }
+
+  // Emit an operand used as an index.
+  LogicalResult emitIndexRel(dxsa::IndexRel index) {
+    Operation *def = index.getOperand().getDefiningOp();
+    if (!def) {
+      return emitError(index.getLoc(), "index must be defined");
+    }
+
+    auto operand = dyn_cast<dxsa::Operand>(*def);
```
asl: You could probably combine the checks into `dyn_cast_if_present`? Or just cast depending on IR invariants. Same everywhere else.

mlir/lib/Target/DXSA/BinaryWriter.cpp
asavonic: Thank you. Only `IntegerType` is possible here, so added `cast<IntegerType>` and an assert to only allow 32- and 64- bit integers.

mlir/lib/Target/DXSA/BinaryWriter.cpp
asavonic: Done.

mlir/lib/Target/DXSA/BinaryWriter.cpp
asavonic: Done. We can just `cast` it.

mlir/lib/Target/DXSA/BinaryWriter.cpp
asavonic: Replaced these errors by assertions.
All instructions are in a builtin module op, which cannot have block arguments.

mlir/lib/Target/DXSA/BinaryWriter.cpp
asavonic: Done.

mlir/lib/Target/DXSA/BinaryWriter.cpp
asavonic: Done.

PR #114, opened 2026-04-28
[mlir][dxsa] Add dcl_temps instruction

tagolog:
Example:
  dxsa.dcl_temps 1


mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -837,6 +842,15 @@ class Parser {
       return builder.buildDclGlobalFlags(DECODE_D3D10_SB_GLOBAL_FLAGS(opcodeToken), loc);
     }
 
+    if (opcode == D3D10_SB_OPCODE_DCL_TEMPS) {
```
asavonic: I think it is better to move all custom instructions to a separate function. This way we can keep `verifyInstructionLength` at the call site and not duplicated it for every instruction. We can also use a `switch` instead of a sequence of `if`.

asavonic (approved):
LGTM overall.

mlir/lib/Target/DXSA/BinaryParser.cpp
tagolog: Done

PR #120, opened 2026-05-01
[mlir][dxsa] Add dcl_indexable_temp instruction

tagolog:
Example:

```mlir
    dxsa.dcl_indexable_temp x<0>[1], 1
    dxsa.dcl_indexable_temp x<1>[23], 2
    dxsa.dcl_indexable_temp x<7>[4096]
```

### DIFF
--- a/mlir/include/mlir/Dialect/DXSA/IR/DXSABitwiseOps.td
+++ b/mlir/include/mlir/Dialect/DXSA/IR/DXSABitwiseOps.td
@@ -1,0 +1,38 @@
+//===- DXSABitwiseOps.td - DXSA bitwise ops -------------*- tablegen -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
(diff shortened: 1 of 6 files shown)


mlir/lib/Dialect/DXSA/IR/DXSA.cpp
```diff
@@ -44,6 +44,54 @@ LogicalResult DclGlobalFlags::verify() {
   return success();
 }
 
+//===----------------------------------------------------------------------===//
+// Custom assembly format helpers
+//===----------------------------------------------------------------------===//
+
+namespace {
+
+/// Parse a register identifier of the form `<prefix><N>` (e.g. `x0`, `r3`).
```
asavonic: We should consider if custom syntax is worth adding at this point, or at all.
It looks good, but we have a lot of instructions to support.

tagolog: Operation format has been refactored to match the latest operand form.

asl (approved):

PR #121, opened 2026-05-01
[mlir][dxsa] Add dcl_thread_group instruction

tagolog:
Example:
  dxsa.dcl_thread_group <x = 1, y = 1, z = 1>

### DIFF
--- a/mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
+++ b/mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
@@ -1322,4 +1322,26 @@ def DXSA_DclResourceRaw : DXSA_Op<"dcl_resource_raw"> {
   let hasVerifier = 1;
 }
 
+def DXSA_DclThreadGroup : DXSA_Op<"dcl_thread_group"> {
+  let summary = "declares compute shader thread group dimensions";
+  let description = [{
+    The `dxsa.dcl_thread_group` operation declares the `$x`, `$y` and `$z`
+    dimensions of a compute shader thread group.
+
+    The product `$x` * `$y` * `$z` must not exceed 1024.
+
+    Example:
+
+    ```mlir
+    dxsa.dcl_thread_group <x = 1, y = 1, z = 1>
+    ```
+  }];
+  let arguments = (ins
+      ConfinedAttr<I32Attr, [IntPositive, IntMaxValue<1024>]>:$x,
+      ConfinedAttr<I32Attr, [IntPositive, IntMaxValue<1024>]>:$y,
+      ConfinedAttr<I32Attr, [IntPositive, IntMaxValue<64>]>:$z);
+  let assemblyFormat = "` ` `<` `x` `=` $x `,` `y` `=` $y `,` `z` `=` $z `>` attr-dict";
+  let hasVerifier = 1;
+}
+
 #endif // DXSA_OPS
--- a/mlir/lib/Dialect/DXSA/IR/DXSA.cpp
+++ b/mlir/lib/Dialect/DXSA/IR/DXSA.cpp
@@ -40,6 +40,18 @@ static void printHexTokens(OpAsmPrinter &printer, Operation *,
                            DenseI32ArrayAttr attr);
 
 //===----------------------------------------------------------------------===//
+// DclThreadGroup
+//===----------------------------------------------------------------------===//
+
+LogicalResult DclThreadGroup::verify() {
+  constexpr int64_t maxTotalThreads = 1024;
+  if (auto total = getX() * getY() * getZ(); total > maxTotalThreads)
+    return emitOpError("thread group size x*y*z must be <= ")
+           << maxTotalThreads << ", got " << total;
+  return success();
+}
+
+//===----------------------------------------------------------------------===//
 // TableGen'd op method definitions
 //===----------------------------------------------------------------------===//
 
--- a/mlir/lib/Target/DXSA/BinaryParser.cpp
+++ b/mlir/lib/Target/DXSA/BinaryParser.cpp
@@ -924,6 +924,13 @@ public:
                                         toAttr(ubound), toAttr(space));
   }
 
+  Instruction buildDclThreadGroup(uint32_t x, uint32_t y, uint32_t z,
+                                  Location loc) {
+    return dxsa::DclThreadGroup::create(
+        builder, loc, builder.getI32IntegerAttr(x),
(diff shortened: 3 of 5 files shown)


mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
```diff
@@ -158,4 +158,14 @@ def DXSA_DclTemps : DXSA_Op<"dcl_temps"> {
   let assemblyFormat = [{ $count attr-dict }];
 }
 
+def DXSA_DclThreadGroup : DXSA_Op<"dcl_thread_group"> {
+  let summary = "declare thread group size";
+  let arguments = (ins
+      ConfinedAttr<I32Attr, [IntPositive, IntMaxValue<1024>]>:$x,
+      ConfinedAttr<I32Attr, [IntPositive, IntMaxValue<1024>]>:$y,
+      ConfinedAttr<I32Attr, [IntPositive, IntMaxValue<64>]>:$z);
+  let assemblyFormat = [{ $x `,` $y `,` $z attr-dict }];
```
asl: bike schedding: maybe it would be better to use syntax outlining the grid better? E.g. `dxsa.dcl_thread_group 8 x 8 x 1`? or even like this: `dxsa.dcl_thread_group <x: 8, y: 8, z: 1>` explicitly showing mesh axis?

mlir/lib/Dialect/DXSA/IR/DXSA.cpp
```diff
@@ -44,6 +44,18 @@ LogicalResult DclGlobalFlags::verify() {
   return success();
 }
 
+//===----------------------------------------------------------------------===//
+// DclThreadGroup
+//===----------------------------------------------------------------------===//
+
+LogicalResult DclThreadGroup::verify() {
+  auto total = uint64_t(getX()) * uint64_t(getY()) * uint64_t(getZ());
```
asl: do you really need cast here? individual operand checks enforces ranges, and 64 * 1024 * 1024 does not overflow 32-bit type.

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -831,6 +838,41 @@ class Parser {
...
+    if (failed(x))
+      return failure();
+    auto y = parseDimension("y", 1024);
+    if (failed(y))
+      return failure();
+    auto z = parseDimension("z", 64);
+    if (failed(z))
+      return failure();
+
+    auto total = uint64_t(*x) * uint64_t(*y) * uint64_t(*z);
```
asl: see note above about cast

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -831,6 +838,41 @@ class Parser {
...
+      if (value == 0 || value > maxValue) {
+        emitError(getLocation(), "thread group ")
+            << dimensionName << " dimension must be in [1, " << maxValue
+            << "], got " << value;
+        return failure();
+      }
+      return value;
+    };
+
+    auto x = parseDimension("x", 1024);
```
asl: I really do not like this kind of repeating pattern, maybe better structure it in a different way, e.g.:
```c++
    auto parseDimension = [&](StringRef dimensionName,
                              uint32_t maxValue, uint32_t &value) -> LogicalResult {
      auto token = parseToken();
      if (failed(token))
        return failure();
      value = *token;
      if (value == 0 || value > maxValue) {
        return emitError(getLocation(), "thread group ")
            << dimensionName << " dimension must be in [1, " << maxValue
            << "], got " << value;
      }
      return success();
    };

    uint32_t x, y, z;
    if (parseDimension("x", 1024, x) || parseDimension("y", 1024, y) || parseDimension("z", 64, z))
      return failure();
    if (auto total = x * y * z; total > 1024) {
       ...
    }
    ...
```

asl: same in other places, btw :)

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
tagolog: I've implemented the operation as the dxsa.dcl_thread_group [8, 8, 1]

mlir/lib/Dialect/DXSA/IR/DXSA.cpp
tagolog: Fixed

mlir/lib/Target/DXSA/BinaryParser.cpp

mlir/lib/Target/DXSA/BinaryParser.cpp

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
```diff
@@ -158,4 +158,24 @@ def DXSA_DclTemps : DXSA_Op<"dcl_temps"> {
...
+    The x and y values must be in [1, 1024], z must be in [1, 64], and the
+    product x * y * z must not exceed 1024.
+
+    Example:
+
+    ```mlir
+    dxsa.dcl_thread_group [8, 8, 1]
+    ```
+  }];
+  let arguments = (ins ConfinedAttr<I64ArrayAttr, [ArrayCount<3>]>:$dims);
```
asl: Maybe better to have `DenseI64ArrayAttr` here? Then you could have direct access to the underlying array w/o need to cast to Attributes for individual entries below.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
tagolog: Finally, I've implemented the operation as
dxsa.dcl_thread_group<x = 1, y = 1, z = 1>
This way the operation is in sync with others.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
tagolog: Finally, I've decide to switch to
dxsa.dcl_thread_group<x = 1, y = 1, z = 1>
without the array.

mlir/lib/Dialect/DXSA/IR/DXSA.cpp
asavonic: Looks like we still cast to `int64_t` in `DclThreadGroup::verify`.

tagolog: Fixed

asl (approved):

PR #122, opened 2026-05-01
[mlir][dxsa] Add dcl_input_control_point_count and dcl_output_control_point_count instructions

tagolog:
Example:
  dxsa.dcl_input_control_point_count 3
  dxsa.dcl_output_control_point_count 4


mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -831,6 +841,32 @@ class Parser {
     return builder.buildDclTemps(count, loc);
   }
 
+  FailureOr<Instruction> parseDclInputControlPointCount(uint32_t opcodeToken,
+                                                        Location loc) {
+    auto count = (opcodeToken & D3D11_SB_INPUT_CONTROL_POINT_COUNT_MASK) >>
```
asl: How's about something along these lines:
```c
#define GET_INST_FIELD(token, INST, FIELD) \
  (((token) & (INST ## _ ## FIELD ## _ ## MASK)) >> (INST ## _ ## FIELD ## _ ## SHIFT))
```

Then you can do:
```c++
auto count = GET_INST_FIELD(opcodeTokem, D3D11_SB_INPUT_CONTROL_POINT, COUNT);
```

Obviously, you can extend down to something like:
```c
#define GET_D3D11_INST_FIELD(token, INST, FIELD) GET_INST_FIELD(token, D3D11 ## _ ## INST, FIELD)
```

asl: Apparently, there is already:
```c
// DECODER MACRO: Given an input control point count declaration token,
// (OpcodeToken0), determine the control point count
#define DECODE_D3D11_SB_INPUT_CONTROL_POINT_COUNT(OpcodeToken0) ((UINT)(((OpcodeToken0)&D3D11_SB_INPUT_CONTROL_POINT_COUNT_MASK)>>D3D11_SB_INPUT_CONTROL_POINT_COUNT_SHIFT))
```

asl (changes requested):
Please use macro from `d3d12TokenizedProgramFormat.hpp`

mlir/lib/Target/DXSA/BinaryParser.cpp

asl (approved):

PR #125, opened 2026-05-07
[mlir][dxsa] Add dcl_tessellator_domain instruction

tagolog:
Example:
  dxsa.dcl_tessellator_domain domain_quad


mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -891,6 +897,14 @@ class Parser {
     return builder.buildDclOutputControlPointCount(count, loc);
   }
 
+  FailureOr<Instruction> parseDclTessellatorDomain(uint32_t opcodeToken,
+                                                   Location loc) {
+    auto domain = DECODE_D3D11_SB_TESS_DOMAIN(opcodeToken);
+    if (domain == D3D11_SB_TESSELLATOR_DOMAIN_UNDEFINED)
```
asavonic: There are only 3 valid values. Perhaps we can check them explicitly here, and reject the rest?

tagolog: Good point. I've switched to the TableGen-generated symbolizeTessellatorDomain instead of a hand-written range check (mirrors SPIR-V approach) https://github.com/access-softek/llvm-project/blob/74e82f733c8b0a94f504bb5e88e219474d01b4b2/mlir/lib/Target/SPIRV/Deserialization/Deserializer.cpp#L162-L164

asavonic (approved):
LGTM.

PR #127, opened 2026-05-08
[mlir][dxsa] Add dcl_tessellator_partitioning instruction

tagolog:
Example:
  dxsa.dcl_tessellator_partitioning partitioning_integer


mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -831,6 +839,17 @@ class Parser {
     return builder.buildDclTemps(count, loc);
   }
 
+  FailureOr<Instruction> parseDclTessellatorPartitioning(uint32_t opcodeToken,
+                                                         Location loc) {
+    auto partitioningMode = DECODE_D3D11_SB_TESS_PARTITIONING(opcodeToken);
+    if (partitioningMode == D3D11_SB_TESSELLATOR_PARTITIONING_UNDEFINED ||
+        partitioningMode > D3D11_SB_TESSELLATOR_PARTITIONING_FRACTIONAL_EVEN) {
+      return emitError(loc, "invalid tessellator partitioning mode");
```
asl: I am wondering, if it would make sense in all these `emitError()` calls everywhere to show the actual invalid value (no need to elaborate on ranges, etc.):
```c++
  return emitError(loc, "invalid tessellator partitioning mode ") << partitioningMode;
```

asavonic (approved):
LGTM.

PR #129, opened 2026-05-08
[mlir][dxsa] Add dcl_input_primitive instruction

tagolog:
Example:
  dxsa.dcl_input_primitive triangle


mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
```diff
@@ -62,6 +62,93 @@ def DXSA_GlobalFlagsAttr :
   let assemblyFormat = "`<` $value `>`";
 }
 
+def DXSA_InputPrimitive_Point        : I32EnumAttrCase<"point",         1>;
+def DXSA_InputPrimitive_Line         : I32EnumAttrCase<"line",          2>;
+def DXSA_InputPrimitive_Triangle     : I32EnumAttrCase<"triangle",      3>;
+def DXSA_InputPrimitive_LineAdj      : I32EnumAttrCase<"line_adj",      6>;
+def DXSA_InputPrimitive_TriangleAdj  : I32EnumAttrCase<"triangle_adj",  7>;
+def DXSA_InputPrimitive_Patch1       : I32EnumAttrCase<"patch1",        8>;
```
asl: Are we using these enum numbers somewhere? Likely not, only textual representation? Maybe you can use tabelgen to generate these things? E.g.
```tablegen
foreach i = 1...32 in
  def DXSA_InputPrimitive_Patch#i : I32EnumAttrCase<"patch" # i, i>;
```
And then continue to other cases?

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
```diff
@@ -62,6 +62,93 @@ def DXSA_GlobalFlagsAttr :
...
+def DXSA_InputPrimitive_Patch32      : I32EnumAttrCase<"patch32",      39>;
+
+def DXSA_InputPrimitive : I32EnumAttr<
+    "InputPrimitive", "geometry shader input primitive type", [
+      DXSA_InputPrimitive_Point,
+      DXSA_InputPrimitive_Line,
+      DXSA_InputPrimitive_Triangle,
+      DXSA_InputPrimitive_LineAdj,
+      DXSA_InputPrimitive_TriangleAdj,
+      DXSA_InputPrimitive_Patch1,
```
asl: Ditto. Use `foreach` to generate range of `PatchN` stuff

tagolog: Fixed

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
tagolog: Good catch on `foreach` - folded both this block and the enum cases list below.
The enum values are used by `BinaryParser.cpp` thus, the loop was implemented to assign value 8 for Patch1
```
foreach i = !range(1, 33) in {
  def DXSA_InputPrimitive_Patch#i :
      I32EnumAttrCase<"patch" # i, !add(i, 7)>;
}
```

asavonic (approved):
LGTM.

PR #130, opened 2026-05-08
[mlir][dxsa] Add AsmPrinter to translate from MLIR to DXSA text

asavonic:
`AsmPrinter` traverses all instructions in an module and prints them in "canonical" assembly syntax. There is no description of its grammar, so the patch follows examples from "Direct3D 11.3 Functional Specification" and output and tests from existing DirectX compiler tools, such as DXC.

Only standard instructions and some basic operand types are supported for now.

### DIFF
--- a/mlir/lib/Target/DXSA/AsmPrinter.cpp
+++ b/mlir/lib/Target/DXSA/AsmPrinter.cpp
@@ -78,11 +78,7 @@ public:
   Printer(raw_ostream &output) : outs(output) { initOpClassMap(opClass); }
 
   LogicalResult emitModule(ModuleOp source) {
-    Region &region = source.getRegion();
-    if (!region.hasOneBlock())
-      return emitError(region.getLoc(), "region should contain only one block");
-
-    for (auto &op : region.front()) {
+    for (auto &op : *source.getBody()) {
       if (auto inst = dyn_cast<dxsa::Instruction>(op)) {
         if (failed(emitInstruction(inst)))
           return failure();
--- a/mlir/lib/Target/DXSA/AsmPrinter.cpp
+++ b/mlir/lib/Target/DXSA/AsmPrinter.cpp
@@ -1,0 +1,362 @@
+#include "mlir/Dialect/DXSA/IR/DXSA.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/Location.h"
+#include "mlir/Target/DXSA/BinaryParser.h"
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/TypeSwitch.h"
+#include "llvm/Support/Debug.h"
+#include "llvm/Support/DebugLog.h"
+#include "llvm/Support/EndianStream.h"
+#include "llvm/Support/LogicalResult.h"
+#include "llvm/Support/NativeFormatting.h"
+
+#include "d3d12TokenizedProgramFormat.hpp"
+
+#define DEBUG_TYPE "export-dxsa-bin"
+
+using namespace mlir;
+using namespace llvm;
+
+// FIXME: remove OpcodeClass from BinaryParser
+enum OpcodeClass {
+  D3D10_SB_FLOAT_OP,
+  D3D10_SB_INT_OP,
+  D3D10_SB_UINT_OP,
+  D3D10_SB_BIT_OP,
+  D3D10_SB_FLOW_OP,
+  D3D10_SB_TEX_OP,
+  D3D10_SB_DCL_OP,
+  D3D11_SB_ATOMIC_OP,
+  D3D11_SB_MEM_OP,
+  D3D11_SB_DOUBLE_OP,
+  D3D11_SB_FLOAT_TO_DOUBLE_OP,
+  D3D11_SB_DOUBLE_TO_FLOAT_OP,
+  D3D11_SB_DEBUG_OP,
+};
+
+using OpClassMap = llvm::DenseMap<StringRef, OpcodeClass>;
+
+static void initOpClassMap(OpClassMap &opcodes) {
+#define SET(OpCode, Name, NumOperands, PrecMask, OpClass)                      \
+  opcodes[Name] = OpClass;
+#include "InstrInfo.def"
+#undef SET
+}
+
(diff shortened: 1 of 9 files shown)


mlir/lib/Target/DXSA/AsmPrinter.cpp
```diff
@@ -0,0 +1,362 @@
...
+  }
+}
+
+class Printer {
+public:
+  Printer(raw_ostream &output) : outs(output) { initOpClassMap(opClass); }
+
+  LogicalResult emitModule(ModuleOp source) {
+    Region &region = source.getRegion();
+    if (!region.hasOneBlock()) {
```
asl: Do we really need this check here? It seems to be something for verifier to enforce.

mlir/lib/Target/DXSA/AsmPrinter.cpp
```diff
@@ -0,0 +1,362 @@
...
+      return emitError(op->getLoc(), "unsupported operand type");
+    }
+
+    bool printSubscript = firstIndexIsSubscript;
+    for (Value value : op.getOperands()) {
+      Operation *index = value.getDefiningOp();
+      assert(index && "undefined index");
+
+      // Non-immediate indices always use subscript syntax.
+      if (!isa<dxsa::IndexImm>(*index)) {
```
asl: Maybe remove extra `{` per code style guide to reduce number of lines?

mlir/lib/Target/DXSA/AsmPrinter.cpp
```diff
@@ -0,0 +1,362 @@
...
+    };
+
+    bool printVec = attr.getNumElements() > 1 || !isInt;
+
+    if (printVec) {
+      outs << "l(";
+    }
+
+    StringRef separator = "";
+    for (APInt v : attr) {
```
asl: no need to copy apint here, grab references.

mlir/lib/Target/DXSA/AsmPrinter.cpp
asavonic: Done.

mlir/lib/Target/DXSA/AsmPrinter.cpp
asavonic: Missed that, thank you!

mlir/lib/Target/DXSA/AsmPrinter.cpp
asl: @asavonic Was this comment forgotten?

mlir/lib/Target/DXSA/AsmPrinter.cpp
```diff
@@ -0,0 +1,347 @@
...
+    return success();
+  }
+
+  // Emit an immediate operand. Its format (integer or floating point) depends
+  // on the instruction it is used for.
+  LogicalResult emitOperandImm(dxsa::OperandImm op, OpcodeClass kind) {
+    auto attr = cast<DenseIntElementsAttr>(op.getImm());
+    auto elementType = cast<IntegerType>(attr.getType().getElementType());
+
+    if (elementType.getWidth() != 32)
```
asl, line 258: Again, this is something verifier should enforce. This is an IR invariant. You should not check it here.

mlir/lib/Target/DXSA/AsmPrinter.cpp
asavonic: Yes. Can you suggest how to add a verifier? ModuleOp is a builtin operand. Do we need to make a dialect-specific operand?

mlir/lib/Target/DXSA/AsmPrinter.cpp
asavonic: Sorry, that should be a FIXME. I don't know how to print 64-bit immediate operands yet, but they are supported in Shader Assembly.

mlir/lib/Target/DXSA/AsmPrinter.cpp
asl: Well, then this is even more redundant :)

`ModuleOp` is single-region single-bb, this is already enforced: https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/IR/BuiltinOps.td#L58

Use `source.getBody()` to get that BB.

Overall, think about IR invariants. Enforcing them in asmprinter is a bad idea. Overall, you should never have `emitError` here unless you have not supported something *yet*.

asavonic: Oh, that is much better, thank you!
Now we only use `emitError` for unimplemented features or as a `Default` case in `TypeSwitch` for unsupported (new?) operations.

mlir/lib/Target/DXSA/AsmPrinter.cpp
```diff
@@ -0,0 +1,344 @@
...
+      break;
+    };
+
+    bool printVec = attr.getNumElements() > 1 || !isInt;
+
+    if (printVec)
+      outs << "l(";
+
+    StringRef separator = "";
+    for (const APInt &v : attr) {
```
asl: Use `interleaveComma` here: https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/ADT/STLExtras.h#L2312

mlir/lib/Target/DXSA/AsmPrinter.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+    return success();
+  }
+
+  LogicalResult emitIndex(Operation *op) {
+    return llvm::TypeSwitch<Operation &, LogicalResult>(*op)
+        .Case<dxsa::IndexImm>(
+            [this](auto index) { return emitIndexImm(index); })
+        .Case<dxsa::IndexRel>(
+            [this](auto index) -> LogicalResult { return emitIndexRel(index); })
```
asl: Why do you need trailing result types here, but not in the first case? Do you need them at all?

mlir/lib/Target/DXSA/AsmPrinter.cpp
asavonic: Done.

mlir/lib/Target/DXSA/AsmPrinter.cpp
asavonic: Thanks! These are leftovers from an early iteration where return type was different. Removed.

PR #132, opened 2026-05-13
[mlir][dxsa] Add dcl_input and dcl_output instructions

tagolog:
Example:
  dxsa.dcl_input <type = input, components = 4, mask = <x>, index = [0]>
  dxsa.dcl_output <type = output, components = 4, mask = <x, y, z, w>, index = [0]>


mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
```diff
@@ -93,6 +95,157 @@ def DXSA_OutputPrimitiveTopologyAttr :
   let assemblyFormat = "$value";
 }
 
+def DXSA_InlineOperandType_Temp                          : I32EnumAttrCase<"temp",                                  0>;
```
asavonic, line 366: I think it would be great if we can use these enums for regular operands as well.

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -941,17 +971,63 @@ class Parser {
...
+  FailureOr<dxsa::InlineOperandAttr> parseInlineOperand() {
+    auto token = parseToken();
+    FAILURE_IF_FAILED(token);
+
+    auto rawOperandType = DECODE_D3D10_SB_OPERAND_TYPE(*token);
+    auto isExtended = DECODE_IS_D3D10_SB_OPERAND_EXTENDED(*token);
+    auto loc = getLocation();
+
+    if (isImmOperand(*token))
+      return emitError(loc, "non-immediate operand is not supported yet");
```
asavonic: Non-immediate or immediate? Now it is inconsistent with `isImmOperand(*token)` condition.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
tagolog: Yes, I plan to refine regular operation right after all DCL operations and reuse this operand for regular operations also.

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -30,6 +30,10 @@ using UINT = unsigned int;
 using namespace mlir;
 using namespace llvm;
 
+#define FAILURE_IF_FAILED(RES)                                                 \
```
asavonic, line 33: How about `RETURN_FAILED`?

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
asavonic: Sounds good. When we use them for regular operations, perhaps we'll need to remove `Inline` prefix, but it is a minor thing that we can address later.

asavonic (approved):
LGTM, thanks!

mlir/lib/Target/DXSA/BinaryParser.cpp
tagolog: FAILURE_IF_FAILED is already used in one of MLIR dialects:

https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/SparseTensor/IR/Detail/DimLvlMapParser.cpp#L15-L18

PR #134, opened 2026-05-22
[mlir][dxsa] Add dcl_function_body and dcl_function_table

asavonic:
DX assembly programs support "virtual" inlining of subroutines by defining a set of functions that might be called for a call site.
The compiler then prepares an inlined specialization for every possible callee. The actual specialization is selected at runtime.

On DXBC level, functions are declared and identified by numbers, grouped into tables, and tables are grouped as interfaces:
```
// Function table for AmbientLight.
dcl_function_body fb0
dcl_function_table ft0 = { fb0 }

// Function table for DirectionalLight.
dcl_function_body fb1
dcl_function_table ft1 = { fb1 }

// main's MyMaterial parameter.
dcl_interface fp0[1][1] = { ft0, ft1 };

// main shader code

// call AmbientLight or DirectionalLight based on function pointer bound
fcall fp0[0][0]
```

This patchset adds only `dcl_function_body` and `dcl_function_table`. Interfaces are going to be in a separate patch.


mlir/test/Target/DXSA/inputs/dcl_function_body.bin
tagolog, line 1: The file content violates the spec.
```hex
90 00 00 00   01 00 00 00
```
should be
```hex
90 00 00 02   01 00 00 00
```
(instruction length in DWORDs including the opcode token)

If the broken length is intentional, I vote to add "_invalid" suffix to the file or even change the file name completely "dcl_instruction_with_invalid_zero_length.bin"

asavonic: Yes, but that is what we have to work with. That opcode token is from [interface1.dxbc](https://github.com/microsoft/DirectXShaderCompiler/blob/main/projects/dxilconv/test/dxbc2dxil/interface1.dxbc) test. I assume the original compiler did not encode length properly for some instructions.

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -1261,7 +1302,11 @@ class Parser {
     if (parseResult.has_value()) {
       if (failed(*parseResult))
         return failure();
-      if (failed(verifyInstructionLength(beginOffset, length)))
+
+      // Some custom instructions are encoded with 0 length, which is
+      // meaningless because the opcode token should be included in
+      // the total length of an instruction.
```
tagolog: Was this change caused by  a "dcl_function_body.bin‎" with a broken instruction length?
If not, can we append a link to the real example of such files in the test suite?

If a real shader is ever found with zero length for some specific opcode, let's apply this bypass for a specific opcode only.

asavonic: I'll add a link to [interface1.dxbc](https://github.com/microsoft/DirectXShaderCompiler/blob/main/projects/dxilconv/test/dxbc2dxil/interface1.dxbc).

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
```diff
@@ -719,4 +719,43 @@ def DXSA_DclOutput : DXSA_Op<"dcl_output"> {
...
+  let description = [{
+
+    The `dxsa.dcl_function_body` declares a unique function body #
+    whose code will appear later in the program at `label fb#`.
+
+    Function bodies are used in `dcl_function_table`.
+
+    Example:
+    ```mlir
+    dxsa.dcl_function_body fb0
```
tagolog: Should be 
dxsa.dcl_function_body 0
(without "fb" prefix)

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
```diff
@@ -719,4 +719,43 @@ def DXSA_DclOutput : DXSA_Op<"dcl_output"> {
   let assemblyFormat = "$operand attr-dict";
 }
 
+def DXSA_DclFunctionBody : DXSA_Op<"dcl_function_body"> {
+  let summary = "declares a function body number";
+  let description = [{
+
+    The `dxsa.dcl_function_body` declares a unique function body #
+    whose code will appear later in the program at `label fb#`.
```
tagolog: Maybe remove references to "fb#" syntax as we are not using it in the mlir for now?

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
```diff
@@ -719,4 +719,43 @@ def DXSA_DclOutput : DXSA_Op<"dcl_output"> {
   let assemblyFormat = "$operand attr-dict";
 }
 
+def DXSA_DclFunctionBody : DXSA_Op<"dcl_function_body"> {
+  let summary = "declares a function body number";
+  let description = [{
+
```
tagolog: Can we remove this empty line? We are not using it for other ops, this is also not a common practice for starting the op description with empty line in the repository.

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -1174,6 +1185,30 @@ class Parser {
...
+  FailureOr<Instruction> parseDclFunctionTable(Location loc) {
+    Token index = parseToken();
+    FAILURE_IF_FAILED(index);
+
+    Token numFunctions = parseToken();
+    FAILURE_IF_FAILED(numFunctions);
+
+    SmallVector<int32_t, 16> functions;
+    functions.resize(*numFunctions);
+    for (uint32_t i = 0; i < numFunctions; ++i) {
```
tagolog:
```suggestion
    for (uint32_t i = 0; i < *numFunctions; ++i) {
```

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -1174,6 +1185,30 @@ class Parser {
...
   }
 
+  FailureOr<Instruction> parseDclFunctionBody(Location loc) {
+    Token index = parseToken();
+    FAILURE_IF_FAILED(index);
+    return builder.buildDclFunctionBody(*index, loc);
+  }
+
+  FailureOr<Instruction> parseDclFunctionTable(Location loc) {
+    Token index = parseToken();
```
tagolog:
```suggestion
    auto index = parseToken();
```

mlir/test/Target/DXSA/dcl_function_body.mlir
```diff
@@ -0,0 +1,6 @@
+// RUN: mlir-translate --import-dxsa-bin %S/inputs/dcl_function_body.bin | FileCheck %s
+
+// CHECK-LABEL: module
+
+// dcl_function_body fb1
```
tagolog: Maybe it is worth no to mix the ops from prev asm with the current MLIR ops? This can be a bit misleading.

tagolog (changes requested):

mlir/test/Target/DXSA/inputs/dcl_function_body.bin
asavonic: Moved this case to `dcl_function_body_invalid_length.bin`.

mlir/lib/Target/DXSA/BinaryParser.cpp
asavonic: Done. So far there is only `dcl_function_body` that we need to handle.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
asavonic: Fixed.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
asavonic: Done.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
asavonic: Done.

mlir/lib/Target/DXSA/BinaryParser.cpp
asavonic: I don't know how this even compiles. Done.

mlir/lib/Target/DXSA/BinaryParser.cpp
asavonic: Changed everywhere.

mlir/test/Target/DXSA/dcl_function_body.mlir
asavonic: Agree. Removed.

mlir/test/Target/DXSA/inputs/dcl_function_body.bin
tagolog: I see the correct instruction length for the function-related operations in the [interface1.dxbc](https://github.com/microsoft/DirectXShaderCompiler/blob/main/projects/dxilconv/test/dxbc2dxil/interface1.dxbc)

```
000002b0: 0100 0000 9000 0002 0000 0000 9000 0002  ................
000002c0: 0100 0000 9100 0004 0000 0000 0100 0000  ................
000002d0: 0000 0000 9100 0004 0100 0000 0100 0000  ................
000002e0: 0100 0000 9208 0006 0000 0000 0100 0000  ................
```

asavonic: You're right. I did not remove the program length token, so it got parsed as a `dcl_function_body`.
Removed this code. Thank you!

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
```diff
@@ -723,203 +719,39 @@ def DXSA_DclOutput : DXSA_Op<"dcl_output"> {
...
-
     ```mlir
-    dxsa.dcl_hs_max_tessfactor 4.200000e+01
+    dxsa.dcl_function_body 0
     ```
   }];
-  let arguments = (ins F32Attr:$max_tessfactor);
-  let assemblyFormat = "$max_tessfactor attr-dict";
-  let hasVerifier = 1;
+  let arguments = (ins I32Attr:$index);
```
tagolog: This argument is positive by design, I vote to restrict it.
```suggestion
  let arguments = (ins ConfinedAttr<I32Attr, [IntNonNegative]>:$index);
```

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
```diff
@@ -723,203 +719,39 @@ def DXSA_DclOutput : DXSA_Op<"dcl_output"> {
...
   }];
-  let arguments = (ins DXSA_InlineOperandAttr:$operand,
-                       ConfinedAttr<I32Attr,
-                           [IntPositive, IntMaxValue<32768>]>:$struct_byte_stride,
-                       ConfinedAttr<I32Attr,
-                           [IntPositive, IntMaxValue<8192>]>:$struct_count);
-  let assemblyFormat =
-      "$operand `,` $struct_byte_stride `,` $struct_count attr-dict";
-  let hasVerifier = 1;
+  let arguments = (ins I32Attr:$index, DenseI32ArrayAttr:$functions);
```
tagolog:
```suggestion
  let arguments = (ins ConfinedAttr<I32Attr, [IntNonNegative]>:$index, DenseI32ArrayAttr:$functions);
```

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -1255,66 +1185,28 @@ class Parser {
...
+    return builder.buildDclFunctionBody(*index, loc);
   }
 
-  FailureOr<Instruction> parseDclHsJoinPhaseInstanceCount(Location loc) {
-    auto count = parseToken();
-    FAILURE_IF_FAILED(count);
-    return builder.buildDclHsJoinPhaseInstanceCount(*count, loc);
-  }
+  FailureOr<Instruction> parseDclFunctionTable(Location loc) {
+    auto index = parseToken();
```
tagolog, line 1195: Extended length bit [31] for `dcl_function_table` is unhandled

The spec says the regular 7-bit length field can overflow, in which case bit [31] is set and the real length lives in the next DWORD:

[d3d12TokenizedProgramFormat.hpp#L2486-L2490](https://github.com/access-softek/llvm-project/blob/0e54a7880591/mlir/lib/Target/DXSA/d3d12TokenizedProgramFormat.hpp#L2486-L2490)
```
// [30:24] Instruction length in DWORDs including the opcode token.
// [31]    0 normally. 1 if extended operand definition, meaning next DWORD
//         contains extended operand description.  If it is extended, then
//         it contains the actual instruction length in DWORDs, since
//         it may not fit into 7 bits if enough functions are defined.
```

DirectXShaderCompiler handles this explicitly [ShaderBinary.cpp#L548-L558](https://github.com/microsoft/DirectXShaderCompiler/blob/main/projects/dxilconv/lib/ShaderBinary/ShaderBinary.cpp#L548-L558)

Right now the parser ignores extended length bit for these opcodes and silently parses the next DWORD as function table identifier.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
asavonic: Thank you. Added restriction.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
asavonic: Done.

mlir/lib/Target/DXSA/BinaryParser.cpp
asavonic: Good catch! Added handling of extended length token.

PR #137, opened 2026-05-25
[mlir][dxsa] Add dcl_output_siv instruction

tagolog:
Example:
  dxsa.dcl_output_siv <type = output, components = 4, mask = <x, y, z, w>, index = [0]>, <position>


asavonic (approved):
Great, thanks!

PR #147, opened 2026-05-28
[mlir][dxsa] Add dcl_resource_raw instruction

tagolog:
Example:
  dxsa.dcl_resource_raw <type = resource, components = 0, index = [0, 0, 3]>, <space = 1>


mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
```diff
@@ -922,4 +922,28 @@ def DXSA_DclTgsmStructured : DXSA_Op<"dcl_tgsm_structured"> {
...
+  let summary = "declares a raw buffer shader input resource bound to a register";
+  let description = [{
+    The `dxsa.dcl_resource_raw` operation declares a raw buffer shader
+    input resource bound to a register.
+
+    Example:
+
+    ```mlir
+    dxsa.dcl_resource_raw <type = resource, components = 0, index = [3]>
+    dxsa.dcl_resource_raw <type = resource, components = 0, index = [0, 0, 3]>, <space = 1>
```
asavonic: Index of `dcl_resource_raw` operand is fake: prior to Shader Model 5.1 it can only be 1D and means the resource number. For Shader Model 5.1 and later, index is always 3D and means [the resource number,  lower bound, upper bound].

Can we define something like this instead?
```
dxsa.dcl_resource_raw <index = 1, lower_bound = 0, upper_bound = 3, space = 1>
```
It's all very confusing, especially since uses of `t#` resource use 2D indices and they work as regular indices [resource number, space index].

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -1317,6 +1323,21 @@ class Parser {
                                           *structCount, loc);
   }
 
+  FailureOr<Instruction> parseDclResourceRaw(Location loc) {
+    auto operand = parseInlineOperand();
+    FAILURE_IF_FAILED(operand);
+
+    std::optional<uint32_t> space;
+    auto indexArray = operand->getIndex();
+    if (indexArray && indexArray.size() == 3) {
```
asavonic: We should emit a diagnostic if index dimension is not 1 or 3.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
tagolog: Fixed. I've used lbound and ubound as in dcl_constant_buffer

mlir/lib/Target/DXSA/BinaryParser.cpp
tagolog: Fixed.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
```diff
@@ -922,4 +922,32 @@ def DXSA_DclTgsmStructured : DXSA_Op<"dcl_tgsm_structured"> {
...
+  }];
+
+  let arguments = (ins
+      I32Attr:$id,
+      OptionalAttr<I32Attr>:$lbound,
+      OptionalAttr<I32Attr>:$ubound,
+      OptionalAttr<I32Attr>:$space);
+  let assemblyFormat = [{
+    ` ` `<` `id` `=` $id
+            (`,` `lbound` `=` $lbound^ `,` `ubound` `=` $ubound
```
asavonic, line 1350: What happens when `lbound` or `ubound` are not set?

tagolog: I've made changes to verify this case.

```mlir
// expected-error@+1 {{'dxsa.dcl_resource_raw' op lbound, ubound and space must be either all set or all absent}}
"dxsa.dcl_resource_raw"() {id = 0 : i32, lbound = 0 : i32} : () -> ()
```

asavonic (approved):
LGTM.

PR #149, opened 2026-05-29
Add dcl_uav_typed, dcl_uav_raw and dcl_uav_structured instructions

tagolog:
Example:
  dxsa.dcl_uav_typed <id = 0, dim = texture2d>,
      <x = float, y = float, z = float, w = float>, <flags = globallyCoherent>
  dxsa.dcl_uav_raw <id = 1>, <flags = globallyCoherent|rasterizerOrdered>
  dxsa.dcl_uav_structured <id = 2, struct_byte_stride = 16>,
      <flags = hasOrderPreservingCounter>


mlir/lib/Dialect/DXSA/IR/DXSA.cpp
```diff
@@ -103,6 +103,56 @@ LogicalResult DclResource::verify() {
   return success();
 }
 
+static LogicalResult verifyUavRange(Operation *op,
+                                    std::optional<uint32_t> lbound,
+                                    std::optional<uint32_t> ubound) {
+  if (lbound && ubound && *lbound > *ubound)
```
asavonic: Need to apply the same changes as in #147. Maybe we can use the same function to verify `lbound` and `ubound`?

tagolog: Fixed

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -937,6 +937,51 @@ class DXBuilder {
                                         toAttr(ubound), toAttr(space));
   }
 
+  dxsa::UAVFlagsAttr buildUavFlagsAttr(std::optional<dxsa::UAVFlags> flags) {
+    return flags ? dxsa::UAVFlagsAttr::get(builder.getContext(), *flags)
```
asavonic: nit:
```suggestion
    return flags ? dxsa::UAVFlagsAttr::get(context, *flags)
```

asavonic (approved):
LGTM.

PR #150, opened 2026-05-29
[Driver][DirectX] Add /Qembed_debug and /Fd flags

kuilpd:


llvm/lib/Target/DirectX/DirectXTargetMachine.cpp
```diff
@@ -54,6 +54,9 @@
 
 using namespace llvm;
 
+extern cl::opt<bool> EmbedDebug;
+extern cl::opt<std::string> PdbDebugPath;
+
```
dzhidzhoev: Do we actually need these lines?

llvm/test/CodeGen/DirectX/ContainerData/DebugName-default-output.test
```diff
@@ -1,17 +0,0 @@
-RUN: llc %S/Inputs/SourceInfo.ll -dx-Zss --filetype=obj -o %t.dxbc
```
dzhidzhoev, line 1: Why is this test getting removed?

llvm/lib/MC/MCDXContainerWriter.cpp
```diff
@@ -132,6 +136,14 @@ ArrayRef<MCDXContainerPart> DXContainerObjectWriter::getParts() {
   return Parts;
 }
 
+bool DXContainerObjectWriter::shouldSkipSection(StringRef SectionName,
+                                                size_t SectionSize) {
+  // Do not write ILDB part if we're not embedding it.
+  if (!EmbedDebug && SectionName == "ILDB")
+    return true;
```
dzhidzhoev, line 143: More sections need to be filtered out (such as VERS). Please refer to the PDB doc.

llvm/lib/Target/DirectX/DXContainerGlobals.cpp
dzhidzhoev: We gotta move this comment closer to 
```c++
  mcdxbc::DebugName DebugName;
  DebugName.setFilename(DebugNameStr);
  SmallString<64> ILDNData;
  raw_svector_ostream OS(ILDNData);
  DebugName.write(OS);
  addSection(M, Globals, ILDNData, "dx.ildn", "ILDN");
```

llvm/lib/Target/DirectX/DXContainerGlobals.cpp
```diff
@@ -168,38 +165,40 @@ void DXContainerGlobals::computeShaderHashAndDebugName(
   if (!MMI.SourceInfo)
     return;
 
-  if (!PdbFileName.empty() && !PdbOutputDir.empty())
-    report_fatal_error(
-        "--dx-pdb-file and --dx-pdb-dir are mutually exclusive options");
+  if (!EmbedDebug && PdbDebugPath.empty())
+    return;
```
dzhidzhoev: ILDN (Shader debug name) part must be emitted when debug info is emitted, regardless of whether `/Zi` or `/Zs` is was used to trigger that.
I would just assume that if we have a `DICompileUnit` in module, then the module is produced by clang in debug info mode, which was triggered by passing `/Zi` or `/Zs` (as the main flags triggering debug info emission). And, in that case, debug name part should be created.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
```diff
@@ -239,19 +239,30 @@ class EmbedDXILPass : public llvm::ModulePass {
     legalizeLifetimeIntrinsics(M);
 
     bool HasDebugInfo = !M.debug_compile_units().empty();
+
+    // Enable EmbedDebug if there is debug info, but it is not being written
+    // to a PDB file.
+    if (HasDebugInfo && !EmbedDebug && PdbDebugPath.empty())
+      EmbedDebug = true;
```
dzhidzhoev, line 248: Modifying a cl::opt does not sound like a good idea to me. These variables are shared with other passes, and they are meant to represent user input. I couldn't find many examples where a pass modifies an option, other than in `SampleProfileLoader` pass initialization.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
```diff
@@ -262,7 +273,7 @@ class EmbedDXILPass : public llvm::ModulePass {
     removeLifetimeIntrinsics(M);
 
     SmallVector<GlobalValue *, 2> Globals;
-    if (HasDebugInfo) {
+    if (HasDebugInfo && (EmbedDebug || !PdbDebugPath.empty())) {
```
dzhidzhoev: Consider simlpifying this check, so that we don't have duplicate complex if conditions. Maybe, `if (!ILDBData.empty())` will be enough?

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
```diff
@@ -148,63 +150,61 @@ static void removeLifetimeIntrinsics(Module &M) {
...
+        }
+        if (NamedMDNode *Args = M.getNamedMetadata("dx.source.args")) {
+          Args->eraseFromParent();
+          M.getOrInsertNamedMetadata("dx.source.args")
+              ->addOperand(MDTuple::get(Ctx, {}));
+        }
+      }
+    } else {
+      // If we don't want debug info, strip it from DXIL.
+      StripDebugInfo(M);
```
dzhidzhoev: StripDebugInfo is an expensive operation (scans all instructions etc). Do we really need to run it if `M.debug_compile_units()` is empty? Asking since HasDebugInfo argument has been removed.

For example, DebugInfo generation on other targets (DwarfDebug or CodeViewDebug) does not run at all if there are no compile units in a module. See the if condition in `DebugHandlerBase::beginModule()`:
```c++
void DebugHandlerBase::beginModule(Module *M) {
  if (M->debug_compile_units().empty())
    Asm = nullptr;
  else
    ...
}
```

I believe we can correctly assume that, without DICompileUnits, no debug info is present at all. Otherwise, LLVM IR is not fully correct (if it has debug info without DICompileUnit, which is normally the root of debug information entries tree).

If it is still necessary to do that, let's make a separate commit which removes extra arguments from writeModule, or refactors the code of it in another way.

clang/test/Driver/dxc_debug.hlsl
dzhidzhoev, line 11: Let's add some checks for command lines created by Driver in presence of /Fd and/or /Qembed_debug. Either here on in a separate test.

llvm/lib/Target/DirectX/DXContainerGlobals.cpp
```diff
@@ -39,15 +39,12 @@ using namespace llvm;
...
-    "dx-pdb-dir",
-    cl::desc("Specify the PDB output directory for DirectX target. The file "
-             "name is derived from the shader hash"),
-    cl::value_desc("directory"));
+extern llvm::cl::opt<bool> EmbedDebug;
+cl::opt<std::string> PdbDebugPath(
+    "dx-Fd",
+    cl::desc("Write debug information to the given file, or automatically "
+             "named file in directory when ending in '\\'"),
+    cl::value_desc("filename"));
```
dzhidzhoev, line 46: @asavonic 

/Fd is an option of original DXC to specify output PDB file name.
Its behavior depends on whether a path ends with path separator:
1. If dxc invocation contains `/Fd=something`, PDB file named "something" is created.
2. If it contains `/Fd=something\` (argument value ends with `\`), DXC attempts to create PDB file named `<MD5-hash-of-bitcode>.pdb` in directory `something\` (`fstat` is not called for the path to determine if it's a directory).

As of my knowledge, llc does not have any existing flag which has similar behavior (changing what we do depending on whether a directory path or a file path is provided as a value for the same argument).
Should we make two options for llc (like `--dx-pdb-dir` and `--dx-pdb-file`), with explicit case distinction, and implement the logic for choosing which llc invocation argument is right in Clang-dxc driver, or should we implement just one llc flag (`--dx-Fd`), and implement the abovementioned DXC logic here in llc?

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
```diff
@@ -239,19 +239,30 @@ class EmbedDXILPass : public llvm::ModulePass {
...
+    if (!HasDebugInfo && EmbedDebug)
+      reportFatalUsageError(
+          "Missing debug info for embedding into the container");
+    // TODO: move this check to DXContainerPDB.cpp when /Zs is implemented.
+    if (!HasDebugInfo && !PdbDebugPath.empty())
+      reportFatalUsageError("Missing debug info for writing to the PDB file");
+
     std::string ILDBData;
-    if (HasDebugInfo) {
+    if (HasDebugInfo && (EmbedDebug || !PdbDebugPath.empty())) {
```
dzhidzhoev: The subcondition `(EmbedDebug || !PdbDebugPath.empty())` is equivalent to `true` at this line, considering this assignment above:
```c++
    if (HasDebugInfo && !EmbedDebug && PdbDebugPath.empty())
      EmbedDebug = true;
```
Assuming HasDebugInfo is true, and execution reached line 255, the body of the if will always be executed.
I'd suggest we create ILDB regardless of flag presence, and do case distinction at later stages, if it is needed.

llvm/lib/Target/DirectX/DXContainerGlobals.cpp
asavonic: Can we always generate `MD5.pdb` file in llc, and let the driver rename it if needed?

llvm/lib/Target/DirectX/DirectXTargetMachine.cpp
kuilpd: No, I forgot to delete these when I experimented with checking these options in this file.

llvm/test/CodeGen/DirectX/ContainerData/DebugName-default-output.test
kuilpd: There is no output to PDB by default anymore, this exact same behavior is tested in `test/CodeGen/DirectX/ContainerData/DebugName-user-directory.test` now.

llvm/lib/MC/MCDXContainerWriter.cpp
kuilpd: VERS is still output there when compiled as a library (which we do in all tests), I'm not sure how to check for this or if we even need to. The other sections are dependent on other flags, as I understand.

llvm/lib/Target/DirectX/DXContainerGlobals.cpp
kuilpd: Will do

llvm/lib/Target/DirectX/DXContainerGlobals.cpp
kuilpd: I wasn't sure where to put this yet, since we don't have a `/Zs` flag yet. This patch implements the logic for existing flags, so this can be changed as needed with `/Zs` patch.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
kuilpd: There is the same check in the driver as well, if `/Qembed_debug` is not specified with `/Zi` present, I enable it for llc with `-dx-embed-debug` anyway, but it does write the warning (same as DXC does). So this check in the backend can only pass when llc is called directly, so I did this just to have the same default behavior. I can remove this, we'll just have to specify the option manually in the tests.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
kuilpd: Will do

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
kuilpd: I joined the 2 arguments because they were the same when called. If `HasDebugInfo` was true, `IsDebug` was passed as true; if `HasDebugInfo` was false, `IsDebug` was passed as false. Basically, `WriteDebug` now implies that there is debug info present.

clang/test/Driver/dxc_debug.hlsl
kuilpd: Alright

llvm/lib/Target/DirectX/DXContainerGlobals.cpp
kuilpd: I implemented this logic in `DXContainerGlobals.cpp:177-185`. I chose to do this in the backend so that we would have the same arguments in both llc and the Driver to avoid confusion, and also so that if we needed a check to see if `Fd` was specified, we wouldn't have to check 2 options all the time.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
kuilpd: Makes sense, will do

llvm/lib/Target/DirectX/DXContainerGlobals.cpp
dzhidzhoev: > Can we always generate `MD5.pdb` file in llc, and let the driver rename it if needed?

Then the driver would need to modify the content of llc's output files, because PDB file name is baked into PDB file and into the main llc output file.

llvm/lib/Target/DirectX/DXContainerGlobals.cpp
dzhidzhoev: Yes, we only have /Zi, as for now.
ILDN must be emitted regardless of whether /Qembed_debug is passed, and regardless of whether /Fd is passed.
Can we simply emit it based on non-emptiness of `M.debug_compile_units()`? It's the indicator of whether /Zi was passed to clang-dxc.

llvm/test/CodeGen/DirectX/ContainerData/DebugName-default-output.test
dzhidzhoev: We will need it for https://github.com/access-softek/llvm-project/pull/150/changes#r3336804179.

llvm/lib/MC/MCDXContainerWriter.cpp
dzhidzhoev: Fair point about VERS.
Yet SRCI should be always filtered out here.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
dzhidzhoev: I believe we can get rid of EmbedDebug usage here and in DXContainerGlobals. Thus, it will only affect MCDXContainerWriter. Then, EmbedDebug can be switched on, when needed, in MCDXContainerWriter. And we won't have to write something to cl::opt from a pass.

llvm/lib/Target/DirectX/DXContainerGlobals.cpp
```diff
@@ -39,15 +39,11 @@ using namespace llvm;
...
-                cl::value_desc("filename"));
-static cl::opt<std::string> PdbOutputDir(
-    "dx-pdb-dir",
-    cl::desc("Specify the PDB output directory for DirectX target. The file "
-             "name is derived from the shader hash"),
-    cl::value_desc("directory"));
+cl::opt<std::string> PdbDebugPath(
+    "dx-Fd",
+    cl::desc("Write debug information to the given file, or automatically "
+             "named file in directory when ending in '\\'"),
```
dzhidzhoev: This description won't work on Unix 😅😅,

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
```diff
@@ -239,6 +241,14 @@ class EmbedDXILPass : public llvm::ModulePass {
     legalizeLifetimeIntrinsics(M);
 
     bool HasDebugInfo = !M.debug_compile_units().empty();
+
+    if (!HasDebugInfo && EmbedDebug)
+      reportFatalUsageError(
+          "Missing debug info for embedding into the container");
+    // TODO: move this check to DXContainerPDB.cpp when /Zs is implemented.
+    if (!HasDebugInfo && !PdbDebugPath.empty())
+      reportFatalUsageError("Missing debug info for writing to the PDB file");
```
dzhidzhoev, line 254: Can we move this check to DXContainerGlobals? And modify
```
  if (M.debug_compile_units().empty())
    return;
```
into something like
```
if (M.debug_compile_units().empty()) {
    if (!PdbDebugPath.empty())
      reportFatalUsageError("Missing debug info for writing to the PDB file");
    return;
}
```

Or is there a problem with such approach?

llvm/test/CodeGen/DirectX/ContainerData/SourceInfo-Args.ll
```diff
@@ -1,4 +1,5 @@
-; RUN: llc %S/Inputs/SourceInfo.ll --filetype=obj -o - | obj2yaml | FileCheck %s --check-prefix=DXC
+; RUN: llc %S/Inputs/SourceInfo.ll --filetype=obj --dx-Fd=%t.pdb
+; RUN: llvm-pdbutil pdb2yaml --dxcontainer %t.pdb | FileCheck %s --check-prefix=DXC
```
dzhidzhoev: Running llc without -o causes SourceInfo.o be created right next to SourceInfo.ll, which is in the source tree, not in the build tree.
We'll need `-o /dev/null` for cases where we don't check main shader output file.

llvm/lib/Target/DirectX/DXContainerGlobals.cpp
kuilpd: Eh, I just copied the exising DXC help. I guess changing it to `/` will at least work on both.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
kuilpd: > Or is there a problem with such approach?

Not really, but I thought it would be better to combine this check with a check for whatever /Zs produces, and do it specifically in DXContainerPDB.cpp since it's PDB-related.

dzhidzhoev: > > Or is there a problem with such approach?
> 
> Not really, but I thought it would be better to combine this check with a check for whatever /Zs produces, and do it specifically in DXContainerPDB.cpp since it's PDB-related.

Currently, DXContainerPDB has one specific input options which determines whether a PDB file should be emitted. It is `dx.pdb.name` metadata node.
Thus, the decision whether PDB file should be emitted is being made in DXContainerGlobals::computeShaderHashAndDebugName. That's where the logic that processes `-dx-Fd` is placed.

By putting the check `if (!PdbDebugPath.empty())` into DXContainerPDB or DXILWriterPass,
1. We scatter the logic of `-dx-Fd` processing across multiple passes.
2. DXContainerPDB will now have two inputs which has overlapping semantics (output file name is determined by `dx.pdb.name`, however, it will still check `-dx-Fd` for some reason, while this flag has already been processed in a previous pass DXContainerGlobals).

That is my concern.

llvm/lib/Target/DirectX/DXContainerGlobals.cpp
```diff
@@ -149,9 +145,9 @@ void DXContainerGlobals::computeShaderHashAndDebugName(
   }
 
   Digest.update(DXILConstant->getRawDataValues());
-  MD5::MD5Result Result = Digest.final();
+  MD5::MD5Result MD5 = Digest.final();
```
dzhidzhoev, line 148: Won't `-fpermissive` complain here that the variable name shadows the namespace name used in the variable type? Should we just leave it as it is?

llvm/test/CodeGen/DirectX/ContainerData/ContainerFlags.ll
```diff
@@ -0,0 +1,34 @@
+;; Check that --dx-embed-debug is enabled by default when debug info is present
```
dzhidzhoev, line 1: Could you add check lines for ILDN part as well?
So that we don't miss it in main container with any combination of options.
And it is in PDB whenever PDB is emitted.

llvm/test/CodeGen/DirectX/ContainerData/DebugName-user-directory.test
```diff
@@ -1,11 +1,13 @@
...
-RUN: llc %S/Inputs/SourceInfo.ll -dx-Zss --filetype=obj -o %t.dxbc --dx-pdb-dir=%t
+RUN: llc %S/Inputs/SourceInfo.ll -dx-Zss --dx-embed-debug --filetype=obj -o %t.dxbc --dx-Fd=%t/
 RUN: llvm-objcopy --dump-section=ILDB=%t0.bc %t.dxbc
 RUN: %md5sum %t0.bc >%t0.bc.md5
 RUN: obj2yaml %t.dxbc >%t.yaml
 RUN: cat %t.yaml %t0.bc.md5 | FileCheck %s --implicit-check-not PDBN --implicit-check-not PDBH
 RUN: %python %S/Inputs/check_pdb_exists.py "%t" "%t0.bc.md5"
 
+CHECK-NOT:   - Name:            PDBN
+CHECK-NOT:   - Name:            PDBH
```
dzhidzhoev: There are `--implicit-check-not`s for these parts at FileCheck call. Is it not enough?

dzhidzhoev (approved):
LGTM with minor comments.

llvm/lib/Target/DirectX/DXContainerGlobals.cpp
kuilpd: Compiled with GCC, works. Afaik it applies only to class and struct names.

llvm/test/CodeGen/DirectX/ContainerData/ContainerFlags.ll
kuilpd: Done

llvm/test/CodeGen/DirectX/ContainerData/DebugName-user-directory.test
kuilpd: Ah, didn't notice them.

PR #153, opened 2026-05-30
[mlir][dxsa] Add dcl_stream instruction

tagolog:
Example:
 dxsa.dcl_stream 0


mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -1144,6 +1149,38 @@ class Parser {
     return builder.buildDclMaxOutputVertexCount(count, loc);
   }
 
+  FailureOr<Instruction> parseDclStream(Location loc) {
+    auto operandToken = parseToken();
+    FAILURE_IF_FAILED(operandToken);
+
+    auto operandType = DECODE_D3D10_SB_OPERAND_TYPE(*operandToken);
```
asavonic: We use `parseInlineOperand` instead of repeating everything here?

tagolog: This is awesome suggeston! Fixed.

asavonic (approved):
LGTM.

PR #156, opened 2026-06-02
[mlir][dxsa] Add unknown instruction

tagolog:
Example:
  dxsa.dcl_temps 1
  dxsa.unknown <tokens = [0x030007FF, 0xDEADBEEF, 0x12345678]>
  dxsa.dcl_temps 2


mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
```diff
@@ -493,6 +493,28 @@ def DXSA_Instruction : DXSA_Op<"instruction"> {
   let assemblyFormat = "$mnemonic $operands attr-dict";
 }
 
+def DXSA_Unknown : DXSA_Op<"unknown"> {
+  let summary = "raw tokens fallback for an undecodable instruction";
+  let description = [{
+    The `dxsa.unknown` operation represents one instruction whose raw
+    tokens could not be decoded into a structured op — unknown opcode,
```
asavonic: Em dash?

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -1531,6 +1571,110 @@ class Parser {
...
+          buffer.begin() + beginOffset + i * tokenSize, endianness::little));
+    return out;
+  }
+
+  /// Run the structured parser over the slice.
+  /// On success the produced ops are kept.
+  /// On failure any partial IR is rewound by guard.
+  bool tryParseAsStructured(const InstructionSlice &slice) {
+    auto expectedEnd = slice.beginOffset + slice.numTokens * tokenSize;
+    DXBuilder::RewindGuard guard(builder);
```
asavonic: If `RewindGuard` is only used once, and we still have to disarm it explicitly, it is not much different than an explicit call to a cleanup routine next to `return false` below.

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -1531,6 +1571,110 @@ class Parser {
                                     loc);
   }
 
+  struct InstructionSlice {
+    size_t beginOffset; // absolute byte offset in `buffer`
+    uint32_t numTokens; // number of tokens to consume (>= 1)
+    StringRef hint;     // short reason, included in the warning
+  };
+
+  std::optional<InstructionSlice> sliceInstruction() {
```
asavonic: `parseInstruction` already decodes and verifies instruction length to verify it. Can we avoid duplicating this code here as well? With custom data it is already non trivial.

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -1531,6 +1571,110 @@ class Parser {
...
+    return InstructionSlice{beginOffset, length, "structured decode failed"};
+  }
+
+  /// Read `numTokens` tokens from `beginOffset` and return them as
+  /// i32s in little-endian order. Caller must guarantee the range is
+  /// in-bounds.
+  SmallVector<int32_t> snapshotTokens(size_t beginOffset, uint32_t numTokens) {
+    SmallVector<int32_t> out(numTokens);
+    for (uint32_t i = 0; i < numTokens; ++i)
+      out[i] = static_cast<int32_t>(support::endian::read<uint32_t>(
```
asavonic: Alternatively, we can reset `currentTokenOffset` to `beginOffset` and use `parseToken` with additional error handling of EOF.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOps.td
tagolog: Fixed.

mlir/lib/Target/DXSA/BinaryParser.cpp
tagolog: RewindGuard removed.

mlir/lib/Target/DXSA/BinaryParser.cpp
tagolog: Refactored completely.

mlir/lib/Target/DXSA/BinaryParser.cpp
tagolog: Good catch! Fixed.

asavonic (approved):
LGTM.

PR #157, opened 2026-06-02
[DirectX][HLSL] Add /Qsource_in_debug_module to llc and Clang DXC driver

dzhidzhoev:
dx.source metadata nodes contain shader source code information, such as source files content, macro definitions used to compile shader, and compiler arguments.

PR for Clang to produce them is currently on review https://github.com/llvm/llvm-project/pull/199689.

Original DXC has two modes to handle these nodes:
1. If /Qsource_in_debug_module is not specified, Source Info part (SRCI) is created inside shader's debug PDB file, and dx.source metadata nodes are replaced with empty nodes.
2. If /Qsource_in_debug_module is specified, dx.source nodes are preserved in debug bitcode module, for debug info consumers to use them directly.

This patch reimplements the behavior described in item 2, and partially reimplements item 1 behavior (SRCI part is not emitted yet).

/Qsource_in_debug_module is added to Clang DXC driver, and the corresponding flag is added to llc. In DXILWriterPass, this flag is used to determine how dx.source metadata nodes should be handled.

### DIFF
--- a/clang/include/clang/Options/Options.td
+++ b/clang/include/clang/Options/Options.td
@@ -9624,6 +9624,11 @@ def dxc_Fc : DXCJoinedOrSeparate<"Fc">,
   HelpText<"Output assembly listing file">;
 def dxc_Frs : DXCJoinedOrSeparate<"Frs">,
   HelpText<"Output additional root signature object file">;
+def dxc_source_in_debug_module
+    : Option<["/", "-"], "Qsource_in_debug_module", KIND_FLAG>,
+      Group<dxc_Group>,
+      Visibility<[DXCOption]>,
+      HelpText<"Embed source code into debug module on DirectX target">;
 def dxil_validator_version : Option<["/", "-"], "validator-version", KIND_SEPARATE>,
   Group<dxc_Group>, Flags<[HelpHidden]>,
   Visibility<[DXCOption, ClangOption, CC1Option]>,
--- a/clang/lib/Driver/ToolChains/Clang.cpp
+++ b/clang/lib/Driver/ToolChains/Clang.cpp
@@ -3832,6 +3832,11 @@ static void RenderHLSLOptions(const ArgList &Args, ArgStringList &CmdArgs,
   if (!Args.hasArg(options::OPT_dxc_no_stdinc) &&
       !Args.hasArg(options::OPT_nostdinc))
     CmdArgs.push_back("-finclude-default-header");
+
+  if (Args.hasArg(options::OPT_dxc_source_in_debug_module)) {
+    CmdArgs.push_back("-mllvm");
+    CmdArgs.push_back("--dx-source-in-debug-module");
+  }
 }
 
 static void RenderOpenACCOptions(const Driver &D, const ArgList &Args,
--- a/llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
+++ b/llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
@@ -28,10 +28,16 @@
 #include "llvm/InitializePasses.h"
 #include "llvm/Pass.h"
 #include "llvm/Support/Alignment.h"
+#include "llvm/Support/CommandLine.h"
 #include "llvm/Transforms/Utils/ModuleUtils.h"
 
 using namespace llvm;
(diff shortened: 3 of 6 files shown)


clang/include/clang/Options/Options.td
```diff
@@ -9624,6 +9624,11 @@ def dxc_Fc : DXCJoinedOrSeparate<"Fc">,
   HelpText<"Output assembly listing file">;
 def dxc_Frs : DXCJoinedOrSeparate<"Frs">,
   HelpText<"Output additional root signature object file">;
+def dxc_source_in_debug_module
+    : Option<["/", "-"], "Qsource_in_debug_module", KIND_FLAG>,
+      Group<dxc_Group>,
+      Visibility<[DXCOption]>,
+      HelpText<"Embed source code into debug module on DirectX target">;
```
kuilpd, line 9631: You can shorten this by using `class DXCFlag` like other flags do.

clang/lib/Driver/ToolChains/Clang.cpp
```diff
@@ -3832,6 +3832,11 @@ static void RenderHLSLOptions(const ArgList &Args, ArgStringList &CmdArgs,
   if (!Args.hasArg(options::OPT_dxc_no_stdinc) &&
       !Args.hasArg(options::OPT_nostdinc))
     CmdArgs.push_back("-finclude-default-header");
+
+  if (Args.hasArg(options::OPT_dxc_source_in_debug_module)) {
+    CmdArgs.push_back("-mllvm");
+    CmdArgs.push_back("--dx-source-in-debug-module");
```
kuilpd, line 3838: Should we use a single dash for all llc options, just for the sake of consistency? I think most of the options are used like that.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
```diff
@@ -155,6 +171,16 @@ class EmbedDXILPass : public llvm::ModulePass {
     // fail the Module Verifier if performed in an earlier pass
     legalizeLifetimeIntrinsics(M);
 
+    // Replace dx.source metadata nodes with stubs.
```
kuilpd, line 174: Is this meant to be merged with the current upstream? I'm guessing it needs backend PDB functionality to fully implement, shouldn't it be added later?

asavonic (approved):
> Original DXC has two modes to handle these nodes:
> 
>   1. If /Qsource_in_debug_module is not specified, Source Info part (SRCI) is created inside shader's debug PDB file, and dx.source metadata nodes are replaced with empty nodes.
> 
>   2. If /Qsource_in_debug_module is specified, dx.source nodes are preserved in debug bitcode module, for debug info consumers to use them directly.
> 
> This patch reimplements the behavior described in item 2.

I suppose it also implements item 1 here: clears dx.* debug metadata, but doesn't generate SRCI yet?

LGTM, thank you!

clang/lib/Driver/ToolChains/Clang.cpp
dzhidzhoev: > Should we use a single dash for all llc options, just for the sake of consistency?

AFAIK most of llc options (which are rather target-specific or pass-specific) use double dash (that's what I see in `llc --help` output):
```
PS D:\llvm\build\native> bin\llc --help | grep -v -- '--' | grep -c '\-\S' (estimate the number of lines with single dash options)
32
PS D:\llvm\build\native> bin\llc --help | grep -c '\--'
219 
```

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
dzhidzhoev: It does not strictly require PDB emission functionality.
It would be nice to merge this on top of ILDB part emission patch, but it's not on review as for now. I can postpone sending this PR to LLVM repo if needed.

clang/lib/Driver/ToolChains/Clang.cpp
kuilpd: Hm, I was judging by how they are usually forwarded in `Clang.cpp`, no idea why is it like that.

PR #158, opened 2026-06-02
[DirectX] Write DXIL with debug info to ILDB part

kuilpd:
Internal review before creating this PR in the upstream.

### DIFF
--- a/llvm/lib/MC/MCDXContainerWriter.cpp
+++ b/llvm/lib/MC/MCDXContainerWriter.cpp
@@ -39,7 +39,7 @@ uint64_t DXContainerObjectWriter::writeObject() {
     PartOffset = alignTo(PartOffset, Align(4ul));
     // The DXIL part also writes a program header, so we need to include its
     // size when computing the offset for a part after the DXIL part.
-    if (Sec.getName() == "DXIL")
+    if (Sec.getName() == "DXIL" || Sec.getName() == "ILDB")
       PartOffset += sizeof(dxbc::ProgramHeader);
   }
   assert(PartOffset < std::numeric_limits<uint32_t>::max() &&
@@ -78,12 +78,12 @@ uint64_t DXContainerObjectWriter::writeObject() {
 
     uint64_t PartSize = SectionSize;
 
-    if (Sec.getName() == "DXIL")
+    if (Sec.getName() == "DXIL" || Sec.getName() == "ILDB")
       PartSize += sizeof(dxbc::ProgramHeader);
     // DXContainer parts should be 4-byte aligned.
     PartSize = alignTo(PartSize, Align(4));
     W.write<uint32_t>(static_cast<uint32_t>(PartSize));
-    if (Sec.getName() == "DXIL") {
+    if (Sec.getName() == "DXIL" || Sec.getName() == "ILDB") {
       dxbc::ProgramHeader Header;
       memset(reinterpret_cast<void *>(&Header), 0, sizeof(dxbc::ProgramHeader));
 
--- a/llvm/lib/MC/MCDXContainerWriter.cpp
+++ b/llvm/lib/MC/MCDXContainerWriter.cpp
@@ -39,7 +39,7 @@ uint64_t DXContainerObjectWriter::writeObject() {
     PartOffset = alignTo(PartOffset, Align(4ul));
     // The DXIL part also writes a program header, so we need to include its
     // size when computing the offset for a part after the DXIL part.
-    if (Sec.getName() == "DXIL" || Sec.getName() == "ILDB")
+    if (dxbc::isProgramPart(Sec.getName()))
       PartOffset += sizeof(dxbc::ProgramHeader);
   }
   assert(PartOffset < std::numeric_limits<uint32_t>::max() &&
@@ -78,12 +78,12 @@ uint64_t DXContainerObjectWriter::writeObject() {
 
     uint64_t PartSize = SectionSize;
 
-    if (Sec.getName() == "DXIL" || Sec.getName() == "ILDB")
+    if (dxbc::isProgramPart(Sec.getName()))
       PartSize += sizeof(dxbc::ProgramHeader);
     // DXContainer parts should be 4-byte aligned.
     PartSize = alignTo(PartSize, Align(4));
     W.write<uint32_t>(static_cast<uint32_t>(PartSize));
-    if (Sec.getName() == "DXIL" || Sec.getName() == "ILDB") {
+    if (dxbc::isProgramPart(Sec.getName())) {
       dxbc::ProgramHeader Header;
       memset(reinterpret_cast<void *>(&Header), 0, sizeof(dxbc::ProgramHeader));
 
--- a/llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
+++ b/llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
@@ -151,12 +151,12 @@ static void replaceNamedMetadataArray(Module &M, StringRef Name,
 }
 
 class EmbedDXILPass : public llvm::ModulePass {
-  std::string writeModule(Module &M, bool HasDebugInfo, bool IsDebug) {
+  std::string writeModule(Module &M, bool HasDebugInfo, bool WriteDebug) {
     std::string Data;
     llvm::raw_string_ostream OS(Data);
 
     if (HasDebugInfo) {
-      if (IsDebug) {
+      if (WriteDebug) {
         // Replace dx.source metadata nodes with stubs.
         // TODO: Add /Qsource_in_debug_module flag to enable/disable this.
         LLVMContext &Ctx = M.getContext();
@@ -188,7 +188,6 @@ class EmbedDXILPass : public llvm::ModulePass {
             NMD.eraseFromParent();
       }
     }
-
     const auto DIMap = DXILDebugInfoPass::run(M);
     WriteDXILToFile(M, OS, DIMap);
     return Data;
@@ -227,13 +226,13 @@ public:
       // Clone the module to avoid alternating it with DebugInfoPass
       // before stripping the debug info later.
       ILDBData =
-          writeModule(*llvm::CloneModule(M), HasDebugInfo, /*IsDebug=*/true);
+          writeModule(*llvm::CloneModule(M), HasDebugInfo, /*WriteDebug=*/true);
     }
 
     // Clone the module to save dx.source metadata nodes from stripping, as they
     // are needed for DXILMetadataAnalysisWrapperPass.
     std::string DXILData =
-        writeModule(*llvm::CloneModule(M), HasDebugInfo, /*IsDebug=*/false);
+        writeModule(*llvm::CloneModule(M), HasDebugInfo, /*WriteDebug=*/false);
 
     // We no longer need lifetime intrinsics after bitcode serialization, so we
     // simply remove them to keep the Module Verifier happy after our
--- a/llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
+++ b/llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
@@ -175,14 +175,12 @@ class EmbedDXILPass : public llvm::ModulePass {
           SmallVector<llvm::Module::ModuleFlagEntry, 4> FlagEntries;
           M.getModuleFlagsMetadata(FlagEntries);
           Flags->eraseFromParent();
-          for (unsigned I : seq(FlagEntries.size())) {
-            llvm::Module::ModuleFlagEntry &Entry = FlagEntries[I];
+          for (llvm::Module::ModuleFlagEntry &Entry : FlagEntries) {
             if (Entry.Key->getString() == "Dwarf Version" ||
                 Entry.Key->getString() == "Debug Info Version") {
               continue;
             }
-            M.addModuleFlag(Entry.Behavior, Entry.Key->getString(),
-                            cast<ConstantAsMetadata>(Entry.Val)->getValue());
+            M.addModuleFlag(Entry.Behavior, Entry.Key->getString(), Entry.Val);
           }
         }
         for (NamedMDNode &NMD : llvm::make_early_inc_range(M.named_metadata()))
--- a/llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
+++ b/llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
@@ -18,6 +18,7 @@
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Analysis/ModuleSummaryAnalysis.h"
 #include "llvm/IR/Constants.h"
+#include "llvm/IR/DebugInfo.h"
 #include "llvm/IR/DerivedTypes.h"
 #include "llvm/IR/GlobalVariable.h"
 #include "llvm/IR/IntrinsicInst.h"
@@ -28,6 +29,7 @@
 #include "llvm/InitializePasses.h"
 #include "llvm/Pass.h"
 #include "llvm/Support/Alignment.h"
+#include "llvm/Transforms/Utils/Cloning.h"
 #include "llvm/Transforms/Utils/ModuleUtils.h"
 
 using namespace llvm;
@@ -138,7 +140,75 @@ static void removeLifetimeIntrinsics(Module &M) {
   }
 }
 
+static void replaceNamedMetadataArray(Module &M, StringRef Name,
+                                      ArrayRef<Metadata *> NewOps) {
+  NamedMDNode *NMD = M.getNamedMetadata(Name);
+  if (!NMD)
+    return;
+  NMD->eraseFromParent();
+  M.getOrInsertNamedMetadata(Name)->addOperand(
+      MDTuple::get(M.getContext(), NewOps));
+}
+
 class EmbedDXILPass : public llvm::ModulePass {
+  std::string writeModule(Module &M, bool HasDebugInfo, bool IsDebug) {
+    std::string Data;
+    llvm::raw_string_ostream OS(Data);
+
+    if (HasDebugInfo) {
+      if (IsDebug) {
+        // Replace dx.source metadata nodes with stubs.
+        // TODO: Add /Qsource_in_debug_module flag to enable/disable this.
+        LLVMContext &Ctx = M.getContext();
+        MDString *EmptyString = MDString::get(Ctx, "");
+        replaceNamedMetadataArray(M, "dx.source.contents",
+                                  {EmptyString, EmptyString});
+        replaceNamedMetadataArray(M, "dx.source.defines", {});
+        replaceNamedMetadataArray(M, "dx.source.mainFileName", {EmptyString});
+        replaceNamedMetadataArray(M, "dx.source.args", {});
+      } else {
+        // If we have an ILDB part, strip DXIL from all debug info.
+        StripDebugInfo(M);
+
+        // Also, manually remove debug version flags and dx.source nodes.
+        if (NamedMDNode *Flags = M.getModuleFlagsMetadata()) {
+          SmallVector<llvm::Module::ModuleFlagEntry, 4> FlagEntries;
+          M.getModuleFlagsMetadata(FlagEntries);
+          Flags->eraseFromParent();
+          for (unsigned I : seq(FlagEntries.size())) {
+            llvm::Module::ModuleFlagEntry &Entry = FlagEntries[I];
+            if (Entry.Key->getString() == "Dwarf Version" ||
+                Entry.Key->getString() == "Debug Info Version") {
+              continue;
+            }
+            M.addModuleFlag(Entry.Behavior, Entry.Key->getString(),
+                            cast<ConstantAsMetadata>(Entry.Val)->getValue());
+          }
+        }
+        for (NamedMDNode &NMD : llvm::make_early_inc_range(M.named_metadata()))
+          if (NMD.getName().starts_with("dx.source"))
+            NMD.eraseFromParent();
+      }
+    }
+
+    const auto DIMap = DXILDebugInfoPass::run(M);
+    WriteDXILToFile(M, OS, DIMap);
+    return Data;
+  }
+
+  GlobalVariable *createSectionGlobal(Module &M, StringRef Data,
+                                      StringRef GlobalName,
+                                      StringRef SectionName) {
+    Constant *ModuleConstant =
+        ConstantDataArray::get(M.getContext(), arrayRefFromStringRef(Data));
+    auto *GV = new llvm::GlobalVariable(M, ModuleConstant->getType(), true,
+                                        GlobalValue::PrivateLinkage,
+                                        ModuleConstant, GlobalName);
+    GV->setSection(SectionName);
+    GV->setAlignment(Align(4));
+    return GV;
+  }
+
 public:
   static char ID; // Pass identification, replacement for typeid
   EmbedDXILPass() : ModulePass(ID) {
@@ -148,29 +218,38 @@ public:
   StringRef getPassName() const override { return "DXIL Embedder"; }
 
   bool runOnModule(Module &M) override {
-    std::string Data;
-    llvm::raw_string_ostream OS(Data);
-
     // Perform late legalization of lifetime intrinsics that would otherwise
     // fail the Module Verifier if performed in an earlier pass
     legalizeLifetimeIntrinsics(M);
 
-    const auto DIMap = DXILDebugInfoPass::run(M);
-    WriteDXILToFile(M, OS, DIMap);
+    bool HasDebugInfo = !M.debug_compile_units().empty();
+    std::string ILDBData;
+    if (HasDebugInfo) {
+      // Write DXIL with debug info to ILDB part.
+      // Clone the module to avoid alternating it with DebugInfoPass
+      // before stripping the debug info later.
+      ILDBData =
+          writeModule(*llvm::CloneModule(M), HasDebugInfo, /*IsDebug=*/true);
+    }
+
+    // Clone the module to save dx.source metadata nodes from stripping, as they
+    // are needed for DXILMetadataAnalysisWrapperPass.
+    std::string DXILData =
+        writeModule(*llvm::CloneModule(M), HasDebugInfo, /*IsDebug=*/false);
 
     // We no longer need lifetime intrinsics after bitcode serialization, so we
     // simply remove them to keep the Module Verifier happy after our
     // not-so-legal legalizations
     removeLifetimeIntrinsics(M);
 
-    Constant *ModuleConstant =
-        ConstantDataArray::get(M.getContext(), arrayRefFromStringRef(Data));
-    auto *GV = new llvm::GlobalVariable(M, ModuleConstant->getType(), true,
-                                        GlobalValue::PrivateLinkage,
-                                        ModuleConstant, "dx.dxil");
-    GV->setSection("DXIL");
-    GV->setAlignment(Align(4));
-    appendToCompilerUsed(M, {GV});
+    SmallVector<GlobalValue *, 2> Globals;
+    if (HasDebugInfo) {
+      // Create a GV after both parts are written, otherwise it gets
+      // added to DXIL when `writeModule` is called the second time.
+      Globals.emplace_back(createSectionGlobal(M, ILDBData, "dx.ildb", "ILDB"));
+    }
+    Globals.emplace_back(createSectionGlobal(M, DXILData, "dx.dxil", "DXIL"));
+    appendToCompilerUsed(M, Globals);
     return true;
   }
 
--- a/llvm/include/llvm/BinaryFormat/DXContainer.h
+++ b/llvm/include/llvm/BinaryFormat/DXContainer.h
@@ -261,6 +261,7 @@ LLVM_ABI PartType parsePartType(StringRef S);
 bool isDebugProgramPart(PartType PT);
 
 const char *getProgramPartName(bool IsDebug);
+bool isProgramPart(StringRef PartName);
 
 struct VertexPSVInfo {
   uint8_t OutputPositionPresent;
--- a/llvm/lib/BinaryFormat/DXContainer.cpp
+++ b/llvm/lib/BinaryFormat/DXContainer.cpp
@@ -118,6 +118,10 @@ const char *dxbc::getProgramPartName(bool IsDebug) {
   return IsDebug ? "ILDB" : "DXIL";
 }
 
+bool dxbc::isProgramPart(StringRef PartName) {
+  return PartName == "DXIL" || PartName == "ILDB";
+}
+
 bool ShaderHash::isPopulated() {
   static uint8_t Zeros[16] = {0};
   return Flags > 0 || 0 != memcmp(&Digest, &Zeros, 16);
--- a/llvm/lib/ObjCopy/DXContainer/DXContainerObjcopy.cpp
+++ b/llvm/lib/ObjCopy/DXContainer/DXContainerObjcopy.cpp
@@ -64,7 +64,7 @@ static Error dumpPartToFile(StringRef PartName, StringRef Filename,
   // documentation on the DXContainer format can be found at
   // https://llvm.org/docs/DirectX/DXContainer.html.
 
-  if (PartName == "DXIL" || PartName == "ILDB" || PartName == "STAT")
+  if (llvm::dxbc::isProgramPart(PartName) || PartName == "STAT")
     Contents = Contents.drop_front(sizeof(llvm::dxbc::ProgramHeader));
   if (Contents.empty())
     return createFileError(Filename, object_error::parse_failed,
--- a/llvm/lib/Target/DirectX/DXContainerGlobals.cpp
+++ b/llvm/lib/Target/DirectX/DXContainerGlobals.cpp
@@ -111,17 +111,13 @@ void DXContainerGlobals::addSection(Module &M,
 
 void DXContainerGlobals::computeShaderHashAndDebugName(
     Module &M, SmallVector<GlobalValue *> &Globals) {
+  // TODO: Add -Zss flag to enable/disable calculating shader hash from ILDB.
   auto *DXILConstant =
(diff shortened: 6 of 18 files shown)


llvm/lib/MC/MCDXContainerWriter.cpp
```diff
@@ -39,7 +39,7 @@ uint64_t DXContainerObjectWriter::writeObject() {
     PartOffset = alignTo(PartOffset, Align(4ul));
     // The DXIL part also writes a program header, so we need to include its
     // size when computing the offset for a part after the DXIL part.
-    if (Sec.getName() == "DXIL")
+    if (Sec.getName() == "DXIL" || Sec.getName() == "ILDB")
```
dzhidzhoev: Could we use `dxbc::isProgramPart()`? If it's not merged into mainline, let's do it as well.

llvm/lib/MC/MCDXContainerWriter.cpp
```diff
@@ -78,12 +78,12 @@ uint64_t DXContainerObjectWriter::writeObject() {
 
     uint64_t PartSize = SectionSize;
 
-    if (Sec.getName() == "DXIL")
+    if (Sec.getName() == "DXIL" || Sec.getName() == "ILDB")
```
dzhidzhoev: Same here.

llvm/lib/MC/MCDXContainerWriter.cpp
```diff
@@ -78,12 +78,12 @@ uint64_t DXContainerObjectWriter::writeObject() {
...
       PartSize += sizeof(dxbc::ProgramHeader);
     // DXContainer parts should be 4-byte aligned.
     PartSize = alignTo(PartSize, Align(4));
     W.write<uint32_t>(static_cast<uint32_t>(PartSize));
-    if (Sec.getName() == "DXIL") {
+    if (Sec.getName() == "DXIL" || Sec.getName() == "ILDB") {
```
dzhidzhoev: Same here.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
```diff
@@ -138,7 +140,75 @@ static void removeLifetimeIntrinsics(Module &M) {
...
+        // If we have an ILDB part, strip DXIL from all debug info.
+        StripDebugInfo(M);
+
+        // Also, manually remove debug version flags and dx.source nodes.
+        if (NamedMDNode *Flags = M.getModuleFlagsMetadata()) {
+          SmallVector<llvm::Module::ModuleFlagEntry, 4> FlagEntries;
+          M.getModuleFlagsMetadata(FlagEntries);
+          Flags->eraseFromParent();
+          for (unsigned I : seq(FlagEntries.size())) {
+            llvm::Module::ModuleFlagEntry &Entry = FlagEntries[I];
```
hvdijk: I may be missing something but why isn't this simply
```suggestion
          for (llvm::Module::ModuleFlagEntry &Entry : FlagEntries) {
```
?

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
```diff
@@ -138,7 +140,75 @@ static void removeLifetimeIntrinsics(Module &M) {
...
+  NamedMDNode *NMD = M.getNamedMetadata(Name);
+  if (!NMD)
+    return;
+  NMD->eraseFromParent();
+  M.getOrInsertNamedMetadata(Name)->addOperand(
+      MDTuple::get(M.getContext(), NewOps));
+}
+
 class EmbedDXILPass : public llvm::ModulePass {
+  std::string writeModule(Module &M, bool HasDebugInfo, bool IsDebug) {
```
dzhidzhoev: Let's rename these variables: `HasDebugInfo` -> `ShouldStripDebug` , `IsDebug` -> `WriteDebugPart`, or something like that. So that their purpose will be less confusing for readers.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
```diff
@@ -138,7 +140,75 @@ static void removeLifetimeIntrinsics(Module &M) {
...
+          M.getModuleFlagsMetadata(FlagEntries);
+          Flags->eraseFromParent();
+          for (unsigned I : seq(FlagEntries.size())) {
+            llvm::Module::ModuleFlagEntry &Entry = FlagEntries[I];
+            if (Entry.Key->getString() == "Dwarf Version" ||
+                Entry.Key->getString() == "Debug Info Version") {
+              continue;
+            }
+            M.addModuleFlag(Entry.Behavior, Entry.Key->getString(),
+                            cast<ConstantAsMetadata>(Entry.Val)->getValue());
```
hvdijk: Do we know that `Entry.Val` is always a `ConstantAsMetadata`?

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
```diff
@@ -138,7 +140,75 @@ static void removeLifetimeIntrinsics(Module &M) {
...
+          }
+        }
+        for (NamedMDNode &NMD : llvm::make_early_inc_range(M.named_metadata()))
+          if (NMD.getName().starts_with("dx.source"))
+            NMD.eraseFromParent();
+      }
+    }
+
+    const auto DIMap = DXILDebugInfoPass::run(M);
```
dzhidzhoev, line 191: Let's move this under under `if (IsDebug)` above, following the email discussion we had (Andrew's email).

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
kuilpd: I've copied this loop from DXC's code, not sure how else to do this.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
```diff
@@ -138,7 +140,75 @@ static void removeLifetimeIntrinsics(Module &M) {
...
+        LLVMContext &Ctx = M.getContext();
+        MDString *EmptyString = MDString::get(Ctx, "");
+        replaceNamedMetadataArray(M, "dx.source.contents",
+                                  {EmptyString, EmptyString});
+        replaceNamedMetadataArray(M, "dx.source.defines", {});
+        replaceNamedMetadataArray(M, "dx.source.mainFileName", {EmptyString});
+        replaceNamedMetadataArray(M, "dx.source.args", {});
+      } else {
+        // If we have an ILDB part, strip DXIL from all debug info.
+        StripDebugInfo(M);
```
hvdijk, line 171: This function returns a `bool` indicating whether any debug info was stripped, we can use this to avoid checking through `debug_compile_units().empty()` and get a more reliable answer.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
hvdijk: `addModuleFlag` has an overload that takes a `Metadata*` which seems like it should work.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
kuilpd: I think we specifically need to strip debug info only if we already know if there is debug info. Is there any other way to know this?

dzhidzhoev (commented):
`llvm/test/CodeGen/DirectX/embed-ildb.ll` from dx-debug has lines for testing dx.source presence in output sections:

```
ILDB-DIS: !dx.source.contents
ILDB-DIS: !dx.source.defines
ILDB-DIS: !dx.source.mainFileName
ILDB-DIS: !dx.source.args
...
+; DXIL-DIS-NOT: !dx.source
```

Could you please import that lines into this PR?
So that there will be test coverage for changes made to `!dx.source` processing in this PR.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
dzhidzhoev: > This function returns a bool indicating whether any debug info was stripped, we can use this to avoid checking through debug_compile_units().empty() and get a more reliable answer.

AFAIK if `!llvm.dbg.cu` is empty, LLVM assumes no debug info is present in the file (see https://github.com/access-softek/llvm-project/pull/150#discussion_r3331129744). Are you aware of cases when we have correct and full LLVM IR with debug info but without compile units?

hvdijk: There is no harm in stripping debug info if there is no debug info, the function will just tell you there was nothing to strip, and it will have done a better search than the current code. So something like  `std::unique_ptr<Module> Clone = llvm::CloneModule(M); bool HaveDebugInfo = llvm::StripDebugInfo(*Clone);`, then write the `DXIL` section from `*Clone` (which assuredly has no debug info, regardless of whether there was originally debug info), then if `HaveDebugInfo` is true, create your second clone and write the `ILDB` section from that.

dzhidzhoev: > There is no harm in stripping debug info if there is no debug info

But it performs an extra pass over the whole module which can be avoided when shader is compiled without debug info.

(Anyway, my opinion on that matter should not block this PR.)

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
kuilpd: The call to `WriteDXILToFile` requires a `DXILDebugInfoMap` argument though, I don't think we can not call it. Doesn't seem to do any harm.

dzhidzhoev: @hvdijk mentioned that we can pass an empty map.

kuilpd: Like this?
```
const DXILDebugInfoMap DIMap;
WriteDXILToFile(M, OS, DIMap);
```
I tried, it crashes everything.

dzhidzhoev: It should be like 
```c++
const DXILDebugInfoMap DIMap;
...
if (IsDebug) {
   ...
   DIMap = DebugInfoPass::run(M);
} else {
   ... // No DebugInfoPass::run(M);
}
...
WriteDXILToFile(M, OS, DIMap);
...
}
```

Does it crash in such case as well?

kuilpd: Yes, that's what I meant. I think it still tries to write something from it in `DXILWriter/DXILBitcodeWriter.cpp`, so we either need a separate interface to write without this pass, or some other constructor that correctly initializes DXILDebugInfoMap so that it doesn't crash the writer afterwards, but this is not for this patch and I don't know if it even matters.

dzhidzhoev: > Yes, that's what I meant. I think it still tries to write something from it in `DXILWriter/DXILBitcodeWriter.cpp`, so we either need a separate interface to write without this pass, or some other constructor that correctly initializes DXILDebugInfoMap so that it doesn't crash the writer afterwards, but this is not for this patch and I don't know if it even matters.

That's interesting. Harald might take a look.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
hvdijk: > But it performs an extra pass over the whole module which can be avoided when shader is compiled without debug info.

But we don't know that. It may be that it's impossible to get into that situation when compiling HLSL (I haven't checked), but we do not require IR to have come from HLSL, our own tests don't do that, and our own tests already test scenarios that we care about that are impossible to come up from compiling HLSL. It's easy to write IR that calls `llvm.dbg.value` without defining `!llvm.dbg.cu`, and it's easy to imagine that some third-party LLVM IR producer actually ends up doing this.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
hvdijk: > I tried, it crashes everything.

I did say I don't actually know if that works, just that if it doesn't work, we should easily be able to make it work. Can you give some details of where and how it crashes?

kuilpd: It crashes with this:
```
llvm/include/llvm/Support/Casting.h:572: decltype(auto) llvm::cast(From *) [To = llvm::ConstantDataArray, From = llvm::Constant]: Assertion `isa<To>(Val) && "cast<Ty>() argument of incompatible type!"' failed.
```
Vlad did some digging, it seems like `StripDebugInfo` doesn't fully remove all debug info from the module, for example a `@llvm.dbg.label` intrinsic was still in DXIL afterwards. Calling the `DebugInfoPass` still transforms the leftover debug info, and without running it (I'm guessing) lowering doesn't know what to do with that debug info.

We don't know how to approach this, could you please investigate? I think we basically need something similar to `DebugInfoPass`, but so that it deletes every debug info it finds instead of transforming it to DXIL equivalents.

I'll have to delay making this PR in the upstream.

hvdijk: > Vlad did some digging, it seems like StripDebugInfo doesn't fully remove all debug info from the module, for example a @llvm.dbg.label intrinsic was still in DXIL afterwards.

Ah, I see. This happens because `DXILPrepareModule` has converted debug info to old-style debug intrinsics, but `StripDebugInfo` assumes the module only contains new-style debug records and no longer checks for debug intrinsics to clean up. We may need to move that `M.convertFromNewDbgValues()` call into `DXILDebugInfoPass` somehow, so that the conversion is delayed until later.

kuilpd: > Ah, I see. This happens because `DXILPrepareModule` has converted debug info to old-style debug intrinsics, but `StripDebugInfo` assumes the module only contains new-style debug records and no longer checks for debug intrinsics to clean up. We may need to move that `M.convertFromNewDbgValues()` call into `DXILDebugInfoPass` somehow, so that the conversion is delayed until later.

Is this something that can be quickly fixed, or should I make this PR upstream as is, and then add this fix later?

hvdijk: I think if it doesn't work because of this (it sounds like even in the version where it doesn't crash, it creates an invalid DXIL section by leaving in debug info), we're better off holding off for a moment on upstreaming it. I'm having a look now at whether this is an easy change.

kuilpd: > I think if it doesn't work because of this (it sounds like even in the version where it doesn't crash, it creates an invalid DXIL section by leaving in debug info), we're better off holding off for a moment on upstreaming it. I'm having a look now at whether this is an easy change.

Thank you!

This doesn't really cause any problems other than the disassembler outputting a warning (because we also remove "Debug Info Version"):
```
dxil-dis.exe: warning: ignoring debug info with an invalid version (0) in dxil.bc
```

dzhidzhoev: > > I think if it doesn't work because of this (it sounds like even in the version where it doesn't crash, it creates an invalid DXIL section by leaving in debug info), we're better off holding off for a moment on upstreaming it. I'm having a look now at whether this is an easy change.
> 
> Thank you!
> 
> This doesn't really cause any problems other than the disassembler outputting a warning (because we also remove "Debug Info Version"):
> 
> ```
> dxil-dis.exe: warning: ignoring debug info with an invalid version (0) in dxil.bc
> ```

Could you share module dump before and after stripping, so that we will see which debug metadata entities are still kept?

kuilpd: This is for `test/tools/dxil-dis/di-label.ll`:
Before stripping:
```
target datalayout = "e-m:e-p:32:32-i1:32-i8:8-i16:16-i32:32-i64:64-f16:16-f32:32-f64:64-n8:16:32:64"
target triple = "dxil-ms-dx"

define void @foo(i32 %i) {
entry:
  tail call void @llvm.dbg.value(metadata i32 %i, i64 0, metadata !12, metadata !14), !dbg !15
  br label %label

label:                                            ; preds = %entry
  ret void
}

; Function Attrs: nounwind readnone
declare void @llvm.dbg.value(metadata, i64, metadata, metadata) #0

; Function Attrs: nounwind readnone
declare void @llvm.dbg.label(metadata) #0

attributes #0 = { nounwind readnone }

!llvm.dbg.cu = !{!0}
!dx.shaderModel = !{!7}
!dx.version = !{!8}
!dx.entryPoints = !{!9}
!llvm.module.flags = !{!10, !11}

!0 = distinct !DICompileUnit(language: DW_LANG_C11, file: !1, isOptimized: false, runtimeVersion: 0, emissionKind: 1, subprograms: !2)
!1 = !DIFile(filename: "label.c", directory: "")
!2 = !{!3}
!3 = !DISubprogram(name: "foo", scope: !1, file: !1, line: 1, type: !4, isLocal: false, isDefinition: true, scopeLine: 1, isOptimized: false, function: void (i32)* @foo, variables: !6)
!4 = !DISubroutineType(types: !5)
!5 = !{null}
!6 = !{}
!7 = !{!"lib", i32 6, i32 3}
!8 = !{i32 1, i32 3}
!9 = !{null, !"", null, null, null}
!10 = !{i32 2, !"Dwarf Version", i32 5}
!11 = !{i32 2, !"Debug Info Version", i32 3}
!12 = !DILocalVariable(tag: DW_TAG_arg_variable, name: "i", arg: 1, scope: !3, file: !1, line: 4, type: !13)
!13 = !DIBasicType(name: "int", size: 32, encoding: DW_ATE_signed)
!14 = !DIExpression()
!15 = !DILocation(line: 1, column: 1, scope: !3)
```

After only `StripDebugInfo(M)` (without manually removing "Debug Info Version"):
```
target datalayout = "e-m:e-p:32:32-i1:32-i8:8-i16:16-i32:32-i64:64-f16:16-f32:32-f64:64-n8:16:32:64"
target triple = "dxil-ms-dx"

define void @foo(i32 %i) {
entry:
  tail call void @llvm.dbg.value(metadata i32 %i, i64 0, metadata !5, metadata !12)
  br label %label

label:                                            ; preds = %entry
  ret void
}

; Function Attrs: nounwind readnone
declare void @llvm.dbg.value(metadata, i64, metadata, metadata) #0

; Function Attrs: nounwind readnone
declare void @llvm.dbg.label(metadata) #0

attributes #0 = { nounwind readnone }

!dx.shaderModel = !{!0}
!dx.version = !{!1}
!dx.entryPoints = !{!2}
!llvm.module.flags = !{!3, !4}

!0 = !{!"lib", i32 6, i32 3}
!1 = !{i32 1, i32 3}
!2 = !{null, !"", null, null, null}
!3 = !{i32 2, !"Dwarf Version", i32 5}
!4 = !{i32 2, !"Debug Info Version", i32 3}
!5 = !DILocalVariable(tag: DW_TAG_arg_variable, name: "i", arg: 1, scope: !6, file: !7, line: 4, type: !11)
!6 = !DISubprogram(name: "foo", scope: !7, file: !7, line: 1, type: !8, isLocal: false, isDefinition: true, scopeLine: 1, isOptimized: false, variables: !10)
!7 = !DIFile(filename: "label.c", directory: "")
!8 = !DISubroutineType(types: !9)
!9 = !{null}
!10 = !{}
!11 = !DIBasicType(name: "int", size: 32, encoding: DW_ATE_signed)
!12 = !DIExpression()
```

Both `@llvm.dbg.label` and `@llvm.dbg.value` are still there.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
dzhidzhoev: > and it's easy to imagine that some third-party LLVM IR producer actually ends up doing this.

Not sure about that when it comes to HLSL, but it definitely shouldn't be like that when the goal is to emit DWARF or CodeView (since their emission will just not run).

The thing is, that if we have no compile units, but still have debug info nodes, it may be an issue somewhere in LLVM IR producers, or in an LLVM IR pass which runs before DXILWriterPass. By calling strip on such IRs, DXILWriterPass will just hide such an issue.

> It's easy to write IR that calls llvm.dbg.value without defining !llvm.dbg.cu

It's not that easy to produce such LLVM IR and get a reasonable result, because
1. If a function has no `#dbg !DISubprogram` attachment, it will cause problems with many passes. Particularly, LiveDebugVariables will drop debug intrinsics from MIR generated for this function. (Yes, MIR is not part of DX pipeline, but it is for most of other targets).
2. Definition DISubprograms must have compile unit field, which is enforced by Verifier.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
hvdijk: Can you try on top of https://github.com/hvdijk/llvm-project/commit/directx-delay-converting-debug-info (which in turn is on top of https://github.com/llvm/llvm-project/pull/198318) ?

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
hvdijk: > The thing is, that if we have no compile units, but still have debug info nodes, it may be an issue somewhere in LLVM IR producers, or in an LLVM IR pass which runs before DXILWriterPass.

It may be, sure, but...

> By calling strip on such IRs, DXILWriterPass will just hide such an issue.

...the issue won't be hidden: it will accurately write a DXIL section without debug info, and an ILDB section with the weird debug info.

I'd be fine with turning this into an assert instead somewhere, but silently leaving debug info in the section that specifically is meant to not have debug info seems wrong, especially if we don't know if there may be legitimate reasons for it that we're just not thinking of now.

> Definition DISubprograms must have compile unit field, which is enforced by Verifier.

Note that I didn't say no `DICompileUnit`, just no `!llvm.dbg.cu` named metadata. There are loads of LLVM tests that have a `DICompileUnit` that is not linked through named metadata, a random example is `llvm/test/Transforms/Coroutines/declare-value.ll`.

dzhidzhoev: > > Definition DISubprograms must have compile unit field, which is enforced by Verifier.
> 
> Note that I didn't say no `DICompileUnit`, just no `!llvm.dbg.cu` named metadata. There are loads of LLVM tests that have a `DICompileUnit` that is not linked through named metadata

Such IRs are definitely incorrect from the Verifier's point of view - see `Verifier::verifyCompileUnits()` - it discards debug info containing compile units that are not listed in `!llvm.dbg.cu`,
For the test you referenced:
```
DICompileUnit not listed in llvm.dbg.cu
!4 = distinct !DICompileUnit(language: DW_LANG_Swift, file: !5, isOptimized: false, runtimeVersion: 0, emissionKind: FullDebug)
llvm-dis.exe: warning: ignoring invalid debug info in tmp.bc
```
Tests from `llvm\test\Transforms` sometimes do not care about that - they run a single pass, not full llc pipeline (and `opt` silences invalid debug info warning). So it may have happened that there they've been unverified.

> By calling strip on such IRs, DXILWriterPass will just hide such an issue.
> 
> ...the issue won't be hidden: it will accurately write a DXIL section without debug info, and an ILDB section with the weird debug info.

It may be easier to spot a problem when debug info section requested by the user is not created than when it is created but later debug info is discarded by IR consumers. Usually, IR loader does not raise an error if debug info is incorrect - it silently strips it (or it raises a warning if user explicitly invokes llvm-dis, but not an error).

> I'd be fine with turning this into an assert instead somewhere, but silently leaving debug info in the section that specifically is meant to not have debug info seems wrong, especially if we don't know if there may be legitimate reasons for it that we're just not thinking of now.
> 
> >

That's definitely worth doing if we leave unconditional Strip call 👍 
We could also do it the other way around: detect debug info presence by checking compile units, and add an `assert(StripDebugInfo(M) == HasDebugInfo && "This module must not contain anything to be stripped")`

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
kuilpd: It worked for `StripDebugInfo`, now the same file above looks like this:
```
target datalayout = "e-m:e-p:32:32-i1:32-i8:8-i16:16-i32:32-i64:64-f16:16-f32:32-f64:64-n8:16:32:64"
target triple = "dxil-ms-dx"

define void @foo(i32 %i) {
entry:
  br label %label

label:                                            ; preds = %entry
  ret void
}

!dx.shaderModel = !{!0}
!dx.version = !{!1}
!dx.entryPoints = !{!2}
!llvm.module.flags = !{!3, !4}

!0 = !{!"lib", i32 6, i32 3}
!1 = !{i32 1, i32 3}
!2 = !{null, !"", null, null, null}
!3 = !{i32 2, !"Dwarf Version", i32 5}
!4 = !{i32 2, !"Debug Info Version", i32 3}
```
And I was able to not call the `DebugInfoPass` after stripping the module.

But now in the unstripped module `@llvm.dbg.value` are printed as `#dbg_value`, failing the test `test/CodeGen/DirectX/debug-info.ll`. Is this better this way maybe? The tests for `dxil-dis` are written using `#dbg_` format as well.

hvdijk: > But now in the unstripped module `@llvm.dbg.value` are printed as `#dbg_value`, failing the test `test/CodeGen/DirectX/debug-info.ll`

That suggests that you cherry-picked only the individual commit that I linked to, rather than checking that out and applying your changes on top of it, is that right? The commit is on top of #198318 because it depends on that, it will do the wrong thing otherwise, it will fail in exactly the way that you are seeing.

kuilpd: Ah I see, I rebased and this works now. So how do we approach merging this to upstream? Do I wait for your PRs to be merged first, or make this PR as is now?

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
kuilpd: So, what's the consensus? Do we change the check `!M.debug_compile_units().empty()` to a call to `StripDebugInfo` with an assert?

dzhidzhoev: I won't be blocking it if we use Strip as the source of truth here, but an extra assert for `!M.debug_compile_units().empty()` value would be nice :)
Motivation: other passes (either DX-specific or generic IR passes) may check compile units list still, to check if debug info should be processed, and they won't be calling StripDebugInfo wherever they want to check for debug info presence. So we want to be sure that both debug info presence indicators converge.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
hvdijk: I'd already created https://github.com/llvm/llvm-project/pull/198317 and https://github.com/llvm/llvm-project/pull/198318 as PRs of branches in the `llvm-project` repository, in accordance with https://llvm.org/docs/GitHub.html#stacked-pull-requests. I've opened the PR for that new commit the same way, https://github.com/llvm/llvm-project/pull/201336, and you can open your PR that same way too. It won't be possible to merge the later PRs until the first PRs are merged, but this allows them to be opened and reviewed already.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
hvdijk: The details of how you do it are up to you, but what matters to me is that if we have debug info, even weird debug info, we don't enter the "no debug info" case, at least in an assertions-enabled build. Checking the result of `StripDebugInfo` seemed to me like the simplest way of achieving that, but any other way that also achieves that same goal works for me.

llvm/lib/Target/DirectX/DXILWriter/DXILWriterPass.cpp
dzhidzhoev: > Ah I see, I rebased and this works now. So how do we approach merging this to upstream? Do I wait for your PRs to be merged first, or make this PR as is now?

Currently, LLVM mainline behavior already diverges from DXC: it emits DXIL part with debug info inside it.
What this PR is trying to achieve is:
1. llc should stop emitting debug info into DXIL part module.
2. llc should emit module with debug info into ILDB (the module that is currently being emitted into DXIL).

PDB creation patch depends on the second change. First change depends on Strip (therefore, depends on @hvdijk changes).

I would decouple the second change from this PR, and send it for review, so that PDB creation patch will have less dependencies/will move quicker.

What do you think about that?

hvdijk: > I would decouple the second change from this PR

That means we would have an intermediate stage where modules are emitted without *any* debug info, neither in the DXIL section nor in the ILDB section. Wouldn't this break the existing `dxil-dis` tests?

dzhidzhoev: > > I would decouple the second change from this PR
> 
> That means we would have an intermediate stage where modules are emitted without _any_ debug info, neither in the DXIL section nor in the ILDB section. Wouldn't this break the existing `dxil-dis` tests?

No, it means we may have an intermediate stage where both parts are emitted with debug info.
One PR should add an extra
`CloneModule(), WriteDXILToFile(...)` call (codepath) which would write to `dx.ildb`.
Another PR should change `dx.dxil` emission codepath into something like
```
CloneModule(...)
StripDebugInfo(...)
...
WriteDXILToFile(...)
```
I think it will take longer to get `dx.dxil` change merged.

dzhidzhoev: > Another PR should change dx.dxil emission codepath into something like
> 
> CloneModule(...)
> StripDebugInfo(...)
> ...
> WriteDXILToFile(...)

We can stack this on top of dx.ildb emission patch as well (it will already depend on some other patches anyway).

hvdijk: > No, it means we may have an intermediate stage where both parts are emitted with debug info.

Ah, I see, that is fine for what gets merged into LLVM `main` but that means the stacked PR approach won't work for the "strip debug info from DXIL" part of it. I think that means we won't be able to create that PR yet to get that reviewed, because it depends on multiple other PRs that are not otherwise related and GitHub does not support that. If it's fine with you to not have that part reviewed just yet, that's also fine with me.

dzhidzhoev: > > No, it means we may have an intermediate stage where both parts are emitted with debug info.
> 
> Ah, I see, that is fine for what gets merged into LLVM `main` but that means the stacked PR approach won't work for the "strip debug info from DXIL" part of it. I think that means we won't be able to create that PR yet to get that reviewed, because it depends on multiple other PRs that are not otherwise related and GitHub does not support that. If it's fine with you to not have that part reviewed just yet, that's also fine with me.

Well, we could try sending `dx.ildb` PR with `!M.debug_compile_units().empty()` check to determine debug info presense (that's what I implied when I told `dx.ildb` change does not depend on Strip changes, sorry), and the assert for checking that "StripDebugInfo() did not find any debug info when there are no compile units" could be a part of dx.dxil change.

Yes, if we choose the approach to check debug info presence mainly with StripDebugInfo, then changes for both sections depend on StripDebugInfo call correctness.

dzhidzhoev: > it depends on multiple other PRs that are not otherwise related and GitHub does not support that

AFAIK if we choose this option from https://llvm.org/docs/GitHub.html#stacked-pull-requests:

> 2. Two PRs with a dependency note
> 
> Create PR_1 for feature_1 and PR_2 for feature_2. In PR_2, include a note in the PR summary indicating that it depends on PR_1 (e.g., “Depends on #PR_1”).
> 
> To make review easier, make it clear which commits are part of the base PR and which are new, e.g. “The first N commits are from the base PR”. This helps reviewers focus only on the incremental changes.

We can make a PR which depends on two other PRs.

kuilpd: When I rebased this PR on your https://github.com/hvdijk/llvm-project/commit/directx-delay-converting-debug-info, it all worked. The only problem is that I had to rebase all other commits that have been merged to main since you created those branches (including some DX patches required for this PR). If you rebase your branches on current upstream main, I will be able to stack this PR on your new [#201336](https://github.com/llvm/llvm-project/pull/201336), and we won't have to split anything.

hvdijk: It needs to be a merge rather than a rebase so that the PRs don't get messed up, but sure, done.

kuilpd: @hvdijk I decided to just upload this PR as is (plus an assert) directly on upstream/main, because we need it ASAP for PDB patches. I made a note, that `StripDebugInfo` doesn't strip all the info, but this will be fixed automatically when your PR that delays converting debug info is merged.

hvdijk: That sounds reasonable to me, but the first two patches are now merged, so it's only https://github.com/llvm/llvm-project/pull/201336 that's left. Who is going to be reviewing your PR? If it's Justin, he might be able to review both your PR and that one at the same time if he knows the dependency.

PR #160, opened 2026-06-03
[mlir][dxsa] Add conditional instructions

asavonic:
Conditional instructions execute only when a condition (zero or non zero) is true for an operand. The patch adds the the following instructions:

- if
- breakc
- continuec
- retc
- discard


tagolog (approved):
I vote to merge it to unblock tests and implement https://github.com/access-softek/llvm-project/issues/171 later

PR #163, opened 2026-06-08
[docs][tools][DirectX] Document DXContainer debug info parts

dzhidzhoev:
This patch adds ILDB, ILDN, VERS, SRCI parts format documentation. It describes how to use LLVM tools to inspect them. Additionally, llvm-pdbutil documentation has been expanded with information on how to inspect PDB files generated by the DirectX compiler.

### DIFF
--- a/llvm/docs/CommandGuide/llvm-pdbutil.rst
+++ b/llvm/docs/CommandGuide/llvm-pdbutil.rst
@@ -40,6 +40,7 @@ in the sections that follow.
     reconstructed.
   * :ref:`merge_subcommand` - Given two PDBs, produce a third PDB that is the
     result of merging the two input PDBs.
+  * :ref:`export_subcommand` - Write the contents of a PDB stream to a file.
 
 .. _pretty_subcommand:
 
@@ -407,6 +408,15 @@ Miscellaneous Options
 
  Dump image section headers.
 
+.. option:: -dxcontainer
+
+ When used with :option:`-streams`, dump a summary of the DXContainer stored in
+ the PDB file's DXContainer stream. Shader companion PDB files produced by the
+ DirectX backend store debug-related container parts in this stream. For
+ example::
+
+   llvm-pdbutil dump --streams --dxcontainer shader.pdb
+
 .. option:: -section-map
 
  Dump section map.
@@ -539,8 +549,22 @@ USAGE: :program:`llvm-pdbutil` pdb2yaml [*options*] <input PDB file>
 Summary
 ^^^^^^^
 
+Produce a YAML description of some or all of a PDB file's contents.
+
 Options
 ^^^^^^^
+
+.. option:: -all
+
+ Dump everything :program:`llvm-pdbutil` knows how to dump, including the
+ DXContainer stream when present.
+
+.. option:: -dxcontainer
+
+ Dump the DXContainer stored in the PDB file's DXContainer stream to YAML.
+ For example::
+
+   llvm-pdbutil pdb2yaml --dxcontainer shader.pdb
 
 .. _yaml2pdb_subcommand:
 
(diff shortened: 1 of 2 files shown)


llvm/docs/CommandGuide/llvm-pdbutil.rst
```diff
@@ -407,6 +408,15 @@ Miscellaneous Options
...
  Dump image section headers.
 
+.. option:: -dxcontainer
+
+ When used with :option:`-streams`, dump a summary of the DXContainer stored in
+ the PDB file's DXContainer stream. Shader companion PDB files produced by the
+ DirectX backend store debug-related container parts in this stream. For
+ example::
+
+   llvm-pdbutil dump --streams --dxcontainer shader.pdb
```
kuilpd: These options work independently, `--streams` prints the list of all PDB streams, `--dxcontainer` tries to print only the DXContainer stream if there is one.

llvm/docs/CommandGuide/llvm-pdbutil.rst
```diff
@@ -539,9 +549,23 @@ USAGE: :program:`llvm-pdbutil` pdb2yaml [*options*] <input PDB file>
...
 
+Produce a YAML description of some or all of a PDB file's contents.
+
 Options
 ^^^^^^^
 
+.. option:: -all
+
+ Dump everything :program:`llvm-pdbutil` knows how to dump, including the
+ DXContainer stream when present.
```
kuilpd: I think we shouldn't mention DXContainer here, maybe just make the description the same as in other commands, "Implies most other options."?

llvm/docs/CommandGuide/llvm-pdbutil.rst
```diff
@@ -565,6 +589,54 @@ Options
...
+
+.. option:: --out=<file>
+
+ The file to write the exported stream data to.
+
+.. option:: --dxcontainer
+
+ Export the DXContainer stored in the PDB file's DXContainer stream. This is
+ the usual way to recover the debug-related container parts from a shader
+ companion PDB file.
```
kuilpd, line 633: Should we add that it's effectively a synonym to `--stream=5`?

llvm/docs/DirectX/DXContainer.rst
```diff
@@ -143,6 +143,270 @@ a 128-bit MD5 hash digest. The flags field can either have the value ``0`` to
...
+module flags removed, and ``dx.source`` metadata nodes are stripped from it.
+Those nodes are preserved in the ILDB module when source is embedded in the
+debug module; otherwise they are replaced with empty placeholder values in the
+ILDB module written to the companion PDB file.
+
+By default, when debug information is present and no companion PDB file is
+requested, the ILDB part is embedded in the main DXContainer output. When a
+companion PDB file is requested, the ILDB part is written to the PDB unless
+llc's ``--dx-embed-debug`` is specified, in which case it is written to both the main
+DXContainer and the PDB.
```
kuilpd, line 164: Should this mention the Driver's flags too? Here and where other options are referenced. I know it's technically `llvm` docs, not `clang`, but where else would we write about driver flags and where to use them?

llvm/docs/CommandGuide/llvm-pdbutil.rst
dzhidzhoev: Thank you! Fixed that.

llvm/docs/CommandGuide/llvm-pdbutil.rst
dzhidzhoev: Fixed. I agree with you.

llvm/docs/CommandGuide/llvm-pdbutil.rst
dzhidzhoev: Added that.

llvm/docs/DirectX/DXContainer.rst
dzhidzhoev: Well, I think it should not mention clang-dxc flags here, because of the reason you mention, and because this doc is rather DXContainer format description (I was in doubt even about adding examples of tool flags, but Andrew asked for that specifically, since tools help to investigate DXContainers).

We could add dxc flags description to some clang document. However, I haven't found where we could add it:
1. AFAIK there are some HLSL-specific documents, but they mention only a couple of dxc flags.
2. Some dxc driver flags appear in `tools/clang/docs/html/ClangCommandLineReference.html`, which is automatically generated from `HelpText<>`  of options described in `clang/include/clang/Options/Options.td`. However, only options usable with `clang` tool are visible in that doc file  (not `clang-dxc`-only options, but those which have `ClangOption` visibility).

I think we should discuss separately where to document clang-dxc options, if we choose to add separate rst documentation for them.

kuilpd: Let's ask people tomorrow how to approach this.

PR #164, opened 2026-06-12
[mlir][dxsa] Refine operands

tagolog:
Example:
  dxsa.add r<0>, r<1>, r<2>
  dxsa.add_sat r<0>, r<1>, r<2>
  dxsa.add r<0>, r<1>, l(1.0, 2.0, 3.0, 4.0)
  dxsa.add r<0>, -r<1>, |r<2>|


mlir/include/mlir/Dialect/DXSA/IR/DXSAOpBase.td
```diff
@@ -0,0 +1,41 @@
...
+//===----------------------------------------------------------------------===//
+
+class DXSA_BinaryOp<string mnemonic> : DXSA_Op<mnemonic> {
+  let arguments = (ins
+      DXSA_DstOperandAttr:$dst,
+      DXSA_SrcOperandAttr:$lhs,
+      DXSA_SrcOperandAttr:$rhs,
+      OptionalAttr<DXSA_ComponentMaskAttr>:$precise);
+  let results = (outs);
+  let assemblyFormat = "$dst `,` $lhs `,` $rhs attr-dict";
```
asl: `assemblyFormat` does not use `$precise`. Unused?

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
```diff
@@ -0,0 +1,212 @@
...
+    if (is64) {
+      outBits = llvm::bit_cast<uint64_t>(floatAttr.getValueAsDouble());
+    } else {
+      auto fValue = static_cast<float>(floatAttr.getValueAsDouble());
+      outBits = llvm::bit_cast<uint32_t>(fValue);
+    }
+    return success();
+  }
+  if (auto intAttr = llvm::dyn_cast<IntegerAttr>(attr)) {
+    outBits = intAttr.getValue().getZExtValue();
```
asl: This asserts in the case of value not fitting into 64 bits.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOperand.td
```diff
@@ -0,0 +1,544 @@
...
+    fields omitted:
+
+    - `[<index>, <index>, ...]` - one `#dxsa.index` per index
+      dimension. A single-entry list omits the brackets.
+    - `none` / `scalar` / `vector` / `reserved` - the operand's
+      `#dxsa.operand_components` count. Omitted when it equals the
+      canonical count for the operand type.
+    - `min16f` / `min2_8f` / `min16i` / `min16u` - the
+      `#dxsa.operand_min_precision` hint.
+    - `<chans>` - positional channel list, parsed as a
```
asavonic: Perhaps rename `<chans>` to `<mask>`?

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+template <typename Float>
+static SmallString<48> floatSpellingForcingDecimalPoint(Float value) {
+  SmallString<48> out;
+  llvm::raw_svector_ostream os(out);
+  os << llvm::format("%g", value);
+  StringRef text(out.data(), out.size());
+  if (text.contains('.'))
+    return out;
+  size_t exponentPos = text.find_first_of("eE");
```
asavonic: I think you can instead use LLVM `write_double` function with `FloatStyle::Fixed` style.

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
```diff
@@ -0,0 +1,212 @@
...
+OperandComponents mlir::dxsa::immComponentsFor(size_t count) {
+  switch (count) {
+  case 0:
+    return OperandComponents::none;
+  case 1:
+    return OperandComponents::scalar;
+  case 4:
+    return OperandComponents::vector;
+  default:
+    return OperandComponents::reserved;
```
asavonic: Do we actually need this? Can we do anything with "reserved" number of components other than to emit an error?

mlir/lib/Dialect/DXSA/IR/DXSAOperand.cpp
```diff
@@ -0,0 +1,847 @@
...
+
+//===----------------------------------------------------------------------===//
+// IndexAttr
+//===----------------------------------------------------------------------===//
+
+LogicalResult IndexAttr::verify(function_ref<InFlightDiagnostic()> emitError,
+                                IntegerAttr imm, SrcOperandAttr relative) {
+  if (!imm && !relative)
+    return emitError()
+           << "index must have at least one of immediate or relative";
```
asavonic: Maybe reword this to `index must be either immediate or relative`?

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
```diff
@@ -0,0 +1,212 @@
...
+using namespace mlir::dxsa;
+
+//===----------------------------------------------------------------------===//
+// Immediate parsing for `l(...)` / `d(...)` payloads
+//===----------------------------------------------------------------------===//
+
+/// Parses one literal element of an `l(...)` / `d(...)` payload. Accepts a
+/// float literal, a decimal integer, or a hex integer, and reinterprets the
+/// value as the raw bit pattern at the requested width.
+static ParseResult parseImmValue(AsmParser &parser, bool is64,
```
asavonic: Do we need floating point immediate here at all? IIRC, a type is not encoded, just bit size.

The original disassembler prints them as integer and floats, but I think it does this by looking at the corresponding instruction. If an immediate operand is used in an FP instruction, then it is printed as FP. I don't think we have to do the same here.

The point is, to avoid dancing around different formats, we can support only integer format (or hex - whatever format MLIR can parse by default).

asavonic (commented):
LGTM overall. With "default" attributes hidden, text format doesn't look too verbose.

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
tagolog: Fixed

mlir/include/mlir/Dialect/DXSA/IR/DXSAOperand.td
```diff
@@ -0,0 +1,544 @@
...
+
+    Example:
+
+    ```mlir
+    <x, y, z, w>
+    <x, x, x, x>
+    <x>
+    ```
+  }];
+  let parameters = (ins ArrayRefParameter<"int64_t">:$channels);
```
asl: why `int64_t`? Can it be, say, `-1`? If not, why not just plain `unsigned?

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
```diff
@@ -0,0 +1,225 @@
...
+    } else {
+      auto fValue = static_cast<float>(floatAttr.getValueAsDouble());
+      bits = llvm::bit_cast<uint32_t>(fValue);
+    }
+    return success();
+  }
+  if (auto intAttr = llvm::dyn_cast<IntegerAttr>(attr)) {
+    auto width = is64 ? 64u : 32u;
+    auto value = intAttr.getValue();
+    if (value.getBitWidth() < width) {
```
asl: Does this mean that it would automatically accept, say `42 : i13` and just drop the type promoting to 32-bit int? Is sign-extension always desired? Say, we're having `0x7FFFFFFF: i31`, you will extend this to `0xFFFFFFFF`. Is this ok?

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
```diff
@@ -0,0 +1,225 @@
...
+    out.push_back('0');
+  } else {
+    out.insert(out.begin() + exponentPos, '0');
+    out.insert(out.begin() + exponentPos, '.');
+  }
+  return out;
+}
+
+template <typename Float, typename UInt>
+static bool floatSpellingReparsesToIdenticalBits(Float value, UInt bits) {
```
asl: This is terrible.... Do we really need this?

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
```diff
@@ -0,0 +1,225 @@
...
+static DecodedImmValue decodeImmValue(Int value) {
+  using UInt = std::make_unsigned_t<Int>;
+  constexpr bool is32 = std::is_same_v<Float, float>;
+  auto bits = static_cast<UInt>(value);
+
+  DecodedImmValue imm;
+  imm.bits = bits;
+  imm.byteWidth = sizeof(UInt);
+
+  // Positive zero is the only value spelled as a bare `0`; every other pattern,
```
asl: Why do you need to guess in the *IR*? This is not in general user-written code. IR should be precise.

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
```diff
@@ -0,0 +1,225 @@
...
+  }
+  os << llvm::format_hex(imm.bits, imm.byteWidth * 2 + 2, true);
+}
+
+//===----------------------------------------------------------------------===//
+// Printing and parsing for immediate-literal operand.
+//===----------------------------------------------------------------------===//
+
+/// Parses the body of a 32-bit immediate-literal operand.
+ParseResult mlir::dxsa::parseImm32Body(AsmParser &parser,
```
asl: Why do you need to export all these functions? Aren't these internal business of the parser?

mlir/lib/Dialect/DXSA/IR/DXSAOperand.cpp
```diff
@@ -0,0 +1,847 @@
...
+  if (keyword == "z")
+    return 2;
+  if (keyword == "w")
+    return 3;
+  return std::nullopt;
+}
+
+/// Parses a single channel keyword `x`, `y`, `z` or `w`.
+static ParseResult parseChannel(AsmParser &parser,
+                                SmallVectorImpl<int64_t> &channels) {
```
asl: Why it takes vectors and pushes to it rather than returning item inderectly? This seems to differ from existing parser API.

mlir/lib/Dialect/DXSA/IR/DXSAOperand.cpp
```diff
@@ -0,0 +1,847 @@
...
+
+static bool isImmediateType(OperandType type) {
+  return type == OperandType::l || type == OperandType::d;
+}
+
+static char charFromChannel(int64_t channel) {
+  return channel == 0 ? 'x' : channel == 1 ? 'y' : channel == 2 ? 'z' : 'w';
+}
+
+static std::optional<int64_t> channelFromKeyword(StringRef keyword) {
```
asl: Do you really need to wrap integer into optional? It seems to be used only once, fold into `parseChannel` and you will be able to return error directly.

asl: Oh, btw, and there is `llvm::StringSwitch` for you.

mlir/lib/Dialect/DXSA/IR/DXSAOperand.cpp
```diff
@@ -0,0 +1,847 @@
...
+  if (imm) {
+    auto type = imm.getType();
+    if (!type.isInteger(32) && !type.isInteger(64))
+      return emitError() << "unsupported index type: " << type;
+  }
+  return success();
+}
+
+/// Parses an index entry that begins with the already-parsed integer `value`,
+/// e.g. `2`, `2 : i64`, `2 : i64 + r<...>`.
```
asl: Do you really allow, say, `i17`? It seems you silently ignore it.

mlir/lib/Dialect/DXSA/IR/DXSAOperand.cpp
```diff
@@ -0,0 +1,847 @@
...
+  SrcOperandAttr relative;
+  if (parser.parseCustomAttributeWithFallback(relative))
+    return {};
+  return parser.getChecked<IndexAttr>(loc, ctx, IntegerAttr(), relative);
+}
+
+void IndexAttr::print(AsmPrinter &printer) const {
+  if (auto imm = getImm()) {
+    printer << imm.getValue();
+    if (imm.getType().isInteger(64))
```
asl, line 408: Aren't you hardcode everything to `i32` during parsing (in `parseIntIndexEntry`)?

mlir/lib/Dialect/DXSA/IR/DXSAOperand.cpp
```diff
@@ -0,0 +1,847 @@
...
+  OperandComponentsAttr components = body.components;
+  if (!components)
+    components =
+        OperandComponentsAttr::get(ctx, defaultComponentsFor(*operandType));
+  ComponentMaskAttr mask = body.mask;
+  if (!mask && components.getValue() == OperandComponents::vector)
+    mask = ComponentMaskAttr::get(ctx, ComponentMask::x | ComponentMask::y |
+                                           ComponentMask::z | ComponentMask::w);
+
+  return parser.getChecked<DstOperandAttr>(loc, ctx, *operandType, components,
```
asl: Parsing is great, but for mere mortals you'd probably add some additional attr builders that would automatically add defaults

mlir/lib/Dialect/DXSA/IR/DXSAOperand.cpp
```diff
@@ -0,0 +1,847 @@
...
+        if (isImmediateType(*type))
+          return parseImmSrcOperand(parser, ctx, loc, *type, modifier);
+
+        SrcOperandBody body;
+        OptionalParseResult bodyResult = parseSrcOperandBody(parser, body);
+        if (bodyResult.has_value() && failed(*bodyResult))
+          return {};
+
+        applySrcOperandDefaults(ctx, *type, body);
+        return parser.getChecked<SrcOperandAttr>(
```
asl, line 979: See comment about about default attribute builders

mlir/test/Target/DXSA/operands/index.mlir
```diff
@@ -0,0 +1,39 @@
...
+
+// -----
+
+// CHECK: dxsa.add r<0>, cb<2 : i64 + r<3, <x>>>, r<2>
+dxsa.add r<0>, cb<2 : i64 + r<3, <x>>>, r<2>
+
+// -----
+
+// CHECK: dxsa.add r<0>, cb<[1 : i64, 2 + r<3, <x>>, r<5, <y>>]>, r<2>
+dxsa.add r<0>, cb<[1 : i64, 2 + r<3, <x>>, r<5, <y>>]>, r<2>
```
asl, line 39: you're allowing other types (e.g. `i17` and `i43`). You need either disallow them or test

mlir/test/Target/DXSA/operands/components.mlir
```diff
@@ -0,0 +1,47 @@
+// RUN: mlir-opt %s -split-input-file | FileCheck %s
```
asl: you should use `--verify-roundtrip` in all dialect syntax / op tests

asl (changes requested):
See all comments above

mlir/include/mlir/Dialect/DXSA/IR/DXSAOperand.td
tagolog: The whole operand documentation has been reafctored.

mlir/lib/Dialect/DXSA/IR/DXSAOperand.cpp
tagolog: Fixed.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOpBase.td
tagolog: Fixed. `precise` is now part of the assembly format as an optional prefix modifier:
```
dxsa.add precise <x, y> r<0>, r<1>, r<2>
```

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
tagolog: FP immediates were removed entirely.

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
tagolog: Refactored.

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
tagolog: Agreed, removed. We now accept only integer/hex immediates and always print hex.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOperand.td
tagolog: Fixed

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
tagolog: Good catch! `parseImmValue` now reads a plain integer token via `parseOptionalInteger` instead of a full attribute: a type annotation is rejected.

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
tagolog: FP logic is removed completely for IR to be precisely clear.

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
tagolog: Agreed, no guessing now. The component mode is derived from `(type, payload length)`, illegal lengths are rejected at parse time, and the verifier enforces consistency, so the IR is always precise.

mlir/lib/Dialect/DXSA/IR/DXSAImmediate.cpp
tagolog: Refactored. The whole immediate code moved back to `DXSAOperand.cpp`.

mlir/lib/Dialect/DXSA/IR/DXSAOperand.cpp
tagolog: Fixed.

mlir/lib/Dialect/DXSA/IR/DXSAOperand.cpp
tagolog: Done. Folded into `parseComponent`, which uses `llvm::StringSwitch<unsigned>` and returns the value via an out-param.

mlir/lib/Dialect/DXSA/IR/DXSAOperand.cpp
tagolog: Refactoired. Only `i32`and `i64` are accepted. Verifier rejects any other width. Added `operands/index_invalid.mlir` proove it.

mlir/lib/Dialect/DXSA/IR/DXSAOperand.cpp
tagolog: Done. Added custom builders for `SrcOperandAttr` and `DstOperandAttr`.

mlir/test/Target/DXSA/operands/index.mlir
tagolog: Disallowed and tested.

mlir/test/Target/DXSA/operands/components.mlir
tagolog: All test support --verify-roundtrip now.

mlir/lib/Dialect/DXSA/IR/DXSA.cpp
```diff
@@ -117,6 +114,25 @@ LogicalResult ModuleOp::verify() {
   return success();
 }
 
+//===----------------------------------------------------------------------===//
+// ShaderVersionAttr
+//===----------------------------------------------------------------------===//
+
+Attribute ShaderVersionAttr::parse(AsmParser &parser, Type) {
+  uint8_t major = 0, minor = 0;
+  if (parser.parseLess() || parser.parseInteger(major) || parser.parseComma() ||
```
asl: Why can't you autogenerate this from tablegen?

mlir/include/mlir/Dialect/DXSA/IR/DXSAOperand.td
```diff
@@ -0,0 +1,600 @@
...
+        "index slot list, one entry per dimension">:$index,
+      AttrParameter<"OperandComponentsAttr",
+        "component-count declaration">:$components,
+      OptionalParameter<"OperandMinPrecisionAttr",
+        "minimum precision hint">:$minPrecision,
+      OptionalParameter<"ComponentMaskAttr",
+        "destination writemask">:$mask);
+  let skipDefaultBuilders = 1;
+  let builders = [
+    AttrBuilder<(ins "OperandType":$type,
```
asl, line 489: Do you want to add `AttrBuilder` with just `type` argument? The argument order looks illogical to me – don't `minPrecision` is "more optional" than mask?  Do we want to have `nullptr` as default values for some arguments?

mlir/include/mlir/Dialect/DXSA/IR/DXSAOperand.td
```diff
@@ -0,0 +1,600 @@
...
+        "neg, abs or abs_neg modifier for the operand"
+        >:$modifier,
+      OptionalParameter<"DenseI32ArrayAttr",
+        "32-bit immediate literal payload (for the `l` operand type)"
+        >:$values32,
+      OptionalParameter<"DenseI64ArrayAttr",
+        "64-bit immediate literal payload (for the `d` operand type)"
+        >:$values64);
+  let builders = [
+    // Non-immediate operand: carries no `l(...)` / `d(...)` payload.
```
asl: See comments above.

mlir/lib/Dialect/DXSA/IR/DXSA.cpp
tagolog: Good point. The shader version is now two ConfinedAttr<I32Attr> on the op, so there's no hand-written parse/print and the range is checked declaratively.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOperand.td
tagolog: Done. Reworked the builders completely based on the statistics from the real test files.

mlir/lib/Dialect/DXSA/IR/DXSAOperand.cpp
tagolog: Fixed.

mlir/lib/Dialect/DXSA/IR/DXSAOperand.cpp
tagolog: Fixed. An index literal now defaults to i32 but is no longer silently truncated: a value that doesn't fit i32 is rejected and must be explicitly written as : i64. Added round-trip and overflow tests.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOperand.td
tagolog: Done. Reworked the imm builders.

asl (approved):
Let's ship it and see what will stick

PR #167, opened 2026-06-15
[mlir][dxsa] Add and instruction

tagolog:
Example:
 dxsa.and r<0>, r<1>, r<2>


mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -1632,6 +1632,20 @@ class Parser {
...
+  template <typename BinOpT>
+  FailureOr<Instruction> decodeBinaryOp(size_t beginOffset, uint32_t length,
+                                        uint32_t preciseMask, Location loc) {
+    auto dst = parseDstOperand();
+    FAILURE_IF_FAILED(dst);
+    auto lhs = parseSrcOperand();
+    FAILURE_IF_FAILED(lhs);
+    auto rhs = parseSrcOperand();
+    FAILURE_IF_FAILED(rhs);
+    if (failed(verifyInstructionLength(beginOffset, length)))
```
asl, line 1687: Don't you want to verify length before parsing the contents?

tagolog: I've appended an additional length check before command parsing. However, the rabbit hole is a bit deeper. I've created a separate issue [Bound operand reads to the instruction window in BinaryParser](https://github.com/access-softek/llvm-project/issues/168) to fix it.

mlir/include/mlir/Dialect/DXSA/IR/DXSABitwiseOps.td
```diff
@@ -0,0 +1,38 @@
...
+//===----------------------------------------------------------------------===//
+// dxsa.and
+//===----------------------------------------------------------------------===//
+
+def DXSA_And : DXSA_BinaryOp<"and"> {
+  let summary = "component-wise logical AND";
+  let description = [{
+    The `dxsa.and` operation computes the component-wise logical AND
+    of each pair of 32-bit values from `$rhs` and `$lhs`.
+    32-bit results placed in `$dst`.
```
hvdijk: Small nit on the wording to match existing style of `add`:
```suggestion
    of each pair of 32-bit values `$dst = $lhs & $rhs`.
```
A wider issue, not just for this PR but also for existing instructions, we don't document which operand is which, I can't tell from the documentation whether `$dst` is the first or the last operand. (I think it's the first?)

tagolog: I've applied your wording suggestion.

On operand order: you're right, `$dst` is the first operand (see DXSA_BinaryOp's assemblyFormat: `$dst`, `$lhs`, `$rhs`). Since none of the existing instructions document this and it affects the whole dialect, I'd rather keep this PR consistent with the current ops and file a separate follow-up issue to document operand order across all instructions.

I've created an issue to handle this https://github.com/access-softek/llvm-project/issues/181

asl (approved):

PR #169, opened 2026-06-16
[clang][Driver][DirectX] Add /Zs flag support

dzhidzhoev:
When DXC is called with /Zs flag, it emits "slim" debug info. It effectively means that ILDB section is omitted from the main DXContainer output and from the output PDB file.

This patch reimplements similar behavior in llc and Clang.

Since /Zs is already used in clang-cl driver, and it has different meaning there, /Zs definition in Options.td was generalized, and some changes were made to how the flag is processed in case of CL mode, to preserve the behavior of the flag.


clang/include/clang/Options/Options.td
```diff
@@ -9328,8 +9328,10 @@ def _SLASH_Zp : CLJoined<"Zp">,
 def _SLASH_Zp_flag : CLFlag<"Zp">,
   HelpText<"Set default maximum struct packing alignment to 1">,
   Alias<fpack_struct_EQ>, AliasArgs<["1"]>;
-def _SLASH_Zs : CLFlag<"Zs">, HelpText<"Run the preprocessor, parser and semantic analysis stages">,
-  Alias<fsyntax_only>;
+def _SLASH_Zs
+    : CLFlag<"Zs", [CLOption, DXCOption]>,
+      HelpText<"Run the preprocessor, parser and semantic analysis stages">,
+      HelpTextForVariants<[DXCOption], "Emit slim PDB debug info">;
```
kuilpd, line 9334: Should we change this existing option? Maybe some builds somewhere use this alias. We can still create a new one with the same console name where other DXC options are, and it will be used only through `clang-dxc`.

clang/lib/Driver/ToolChains/Clang.cpp
```diff
@@ -4588,6 +4601,8 @@ renderDebugOptions(const ToolChain &TC, const Driver &D, const llvm::Triple &T,
            SplitDWARFInlining))
         DwarfFission = DwarfFissionKind::None;
     }
+  } else if (Args.hasArg(options::OPT__SLASH_Zs) && D.IsDXCMode()) {
+    DebugInfoKind = llvm::codegenoptions::DebugInfoConstructor;
```
kuilpd, line 4609: I don't see DXC doing the something similar, why are we limiting the debug info? As I understood, the point of /Zs is just to omit ILDB from the PDB file.

clang/lib/Driver/Driver.cpp
```diff
@@ -330,6 +330,11 @@ InputArgList Driver::ParseArgStrings(ArrayRef<const char *> ArgStrings,
   return Args;
 }
 
+static bool isCLSyntaxOnly(const Driver &D, const llvm::opt::ArgList &Args) {
+  return Args.hasArg(options::OPT_fsyntax_only) ||
+         (D.IsCLMode() && Args.hasArg(options::OPT__SLASH_Zs));
+}
+
```
kuilpd, line 337: I'm guessing we wouldn't have to do this if we create a separate DXC-only /Zs option.

llvm/test/CodeGen/DirectX/ContainerData/ContainerFlags.ll
```diff
@@ -31,6 +31,26 @@
...
+;; Check that slim debug (-dx-Zs) omits ILDB from container and companion PDB
+; RUN: llc %S/Inputs/SourceInfo.ll --filetype=obj -dx-Zs --dx-Fd=%t.pdb -o %t.cso
+; RUN: obj2yaml %t.cso | FileCheck %s --check-prefix=SLIM --implicit-check-not=ILDB
+; RUN: llvm-pdbutil pdb2yaml --dxcontainer %t.pdb | FileCheck %s --check-prefix=SLIM --implicit-check-not=ILDB
+
+; SLIM: Parts:
+; SLIM:   - Name:             HASH
+; SLIM:   - Name:             ILDN
+
+;; Check that /Zss can still hash from ILDB when /Zs omits it from output
```
kuilpd, line 43: Does this work fine if we enable `codegenoptions::DebugInfoConstructor` for /Zs? Wouldn't this produce a different hash compared to /Zi?

PR #175, opened 2026-06-17
[mlir][dxsa] Report decode reason in unknown fallback

tagolog:
Example:
 warning: treating next 7 token(s) as unknown: unknown operand type: 255


mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -2206,12 +2218,15 @@ class Parser {
...
-    if (tryParseInstructionOrRewind(instructionLengthInTokens))
+    std::string errorMessage;
+    if (tryParseInstructionOrRewind(instructionLengthInTokens, errorMessage))
       return success();
 
     currentTokenOffset = beginOffset;
-    emitWarning(getLocation()) << "treating next " << instructionLengthInTokens
-                               << " token(s) as unknown";
+    emitWarning(getLocation())
+        << "treating next " << instructionLengthInTokens
```
asavonic, line 2505: Is there a test for these warnings?

asavonic (commented):
Looks good.

mlir/lib/Target/DXSA/BinaryParser.cpp
tagolog: Appended diag rows to the test:
```mlir
// DIAG: warning: treating next 3 token(s) as unknown: instruction length mismatch
```

asavonic (approved):
Great, thank you!

PR #178, opened 2026-06-18
[mlir][dxsa] Add mad, min, max, mul and sincos instructions

tagolog:
Example:
 dxsa.mad r<0>, r<1>, r<2>, r<3>
 dxsa.sincos null, r<1>, -r<2>


mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -786,16 +786,36 @@ class DXBuilder {
...
-    return OpT::create(builder, loc, dst, lhs, rhs, preciseAttr);
+    return OpT::create(builder, loc, dst, lhs, rhs,
+                       buildPreciseAttr(preciseMask));
+  }
+
+  template <typename OpT>
+  Instruction
+  buildMultiplyAddOp(dxsa::DstOperandAttr dst, dxsa::SrcOperandAttr lhs,
+                     dxsa::SrcOperandAttr rhs, dxsa::SrcOperandAttr acc,
+                     uint32_t preciseMask, Location loc) {
```
asl: I'd rename it to `buildTernaryOp`. After all, there are 3 operands and nothing madd-specific.

tagolog: Done: dropped the arity from the names entirely. Instead of `buildBinaryOp`, `buildMultiplyAddOp`, `buildSincosOp` plus three `decodeSaturable*Op`, there's now a single `decodeSaturableOp<OpT, OpSatT, NumDst, NumSrc>` parametrized by the dst/src operand counts. So the "ternary" naming question is moot now.

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -1649,22 +1660,33 @@ class Parser {
...
-                          uint32_t preciseMask, Location loc) {
-    auto dst = parseDstOperand();
-    FAILURE_IF_FAILED(dst);
-    auto lhs = parseSrcOperand();
-    FAILURE_IF_FAILED(lhs);
-    auto rhs = parseSrcOperand();
-    FAILURE_IF_FAILED(rhs);
+  template <std::size_t N, typename OperandT>
+  FailureOr<std::array<OperandT, N>>
+  parseOperands(FailureOr<OperandT> (Parser::*parseOperand)()) {
```
asl:
```suggestion
  parseNOperands(FailureOr<OperandT> (Parser::*parseOperand)()) {
```

tagolog: Good point! Done.

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -786,27 +788,25 @@ class DXBuilder {
...
+  Instruction buildOpWithPreciseMask(
+      uint32_t preciseMask, Location loc,
+      const std::array<dxsa::DstOperandAttr, NumDstOperands> &dsts,
+      const std::array<dxsa::SrcOperandAttr, NumSrcOperands> &srcs) {
+    return std::apply(
+        [&](auto... operands) {
+          return OpT::create(builder, loc, operands...,
+                             buildPreciseAttr(preciseMask));
+        },
+        std::tuple_cat(dsts, srcs));
```
hvdijk: Comment copied from another PR: `std::tuple_cat` isn't required to work with `std::array` in C++17, and LLVM only requires C++17

tagolog: Good catch, thanks! I've switched to nested std::apply, which is guaranteed to work with std::array.

asl (approved):

hvdijk (approved):
Looks good, thanks. There might possibly be further improvements we could do, not sure, but it seems better to just get this merged first.

PR #187, opened 2026-06-19
[mlir][dxsa] Add dtof, ftod, dtoi, dtou, itod and utod instructions

tagolog:
Example:
  dxsa.dtof r<0, <x, y>>, r<1>
  dxsa.ftod r<0>, r<1>
  dxsa.dtoi r<0, <x, y>>, r<1>
  dxsa.dtou r<0, <x, y>>, r<1>
  dxsa.itod r<0>, r<1>
  dxsa.utod r<0>, r<1>


mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -2276,6 +2276,18 @@ class Parser {
...
+    case D3D11_SB_OPCODE_FTOD:
+      return UNARY_OP(Ftod);
+    case D3D11_1_SB_OPCODE_DTOI:
+      return UNARY_OP(Dtoi);
+    case D3D11_1_SB_OPCODE_DTOU:
+      return UNARY_OP(Dtou);
+    case D3D11_1_SB_OPCODE_ITOD:
+      return UNARY_OP(Itod);
+    case D3D11_1_SB_OPCODE_UTOD:
+      return UNARY_OP(Utod);
```
hvdijk: We really should have some sort of order for the instructions in this file. I have, for lack of a better option, just kept the bitwise instructions I've added in alphabetical order; it's going to cause merge conflicts to have other unrelated instructions inserted in the middle of them when I add more. Do you have any particular preference on this?

mlir/include/mlir/Dialect/DXSA/IR/DXSADoubleTypeConversionOps.td
```diff
@@ -0,0 +1,175 @@
...
+// dxsa.dtou
+//===----------------------------------------------------------------------===//
+
+def DXSA_Dtou : DXSA_UnaryOp<"dtou"> {
+  let summary = "component-wise double-precision float to unsigned int32 conversion";
+  let description = [{
+    The `dxsa.dtou` operation converts each double-precision component of
+    `$src` to an unsigned 32-bit integer using truncation (round toward zero)
+    and writes the result to `$dst`. Inputs are clamped to
+    [0.0, 4294967295.999] before conversion; NaN inputs produce zero.
```
hvdijk: I see that this is copied from the Microsoft documentation, but it doesn't make sense? `4294967295.999` isn't representable in double precision floating point, but clamping to `[0.0, x]` for any `4294967295.0 <= x < 4294967296.0` produces identical results, so it won't actually clamp to  `4294967295.999` but also it doesn't make any difference what it actually clamps to, the decimals are just arbitrary. Do we want to just copy the documentation, or do we want to clear it up?

tagolog: Good point. I've dropped the clamp/NaN wording from both dtoi and dtou so they now just state "rounding toward zero", matching the existing ftoi/ftou descriptions which omit it too.

mlir/lib/Target/DXSA/BinaryParser.cpp
tagolog: Good point. I've reworked this: the conversion opcodes now live together in a dedicated "Type conversion instructions" block in the main switch, kept in alphabetical order (dtof, dtoi, dtou, f16tof32, f32tof16, ftod, ftoi, ftou, itod, itof, utod, utof), so they no longer get inserted into the middle of your bitwise block.

mlir/include/mlir/Dialect/DXSA/IR/DXSADoubleTypeConversionOps.td
hvdijk: I think even though the wording didn't quite make sense, it was intended to express an important thing: it's well-defined to convert a `double` that is not in the range of `unsigned` using `dtou`, and likewise for the other ops. With LLVM's [`fptoui`](https://llvm.org/docs/LangRef.html#fptoui-to-instruction), this would be explicitly undefined, so highlighting that `dtou` is different seems worthwhile. But I'm not sure how to best express that. It's fine to say, IMO, that based on `ftou`'s description also missing this we leave this as an issue to be resolved later; if so, shall I open an issue to make sure it's tracked (even if we never get to it)?

hvdijk (approved):
Looks good to me, thanks. The same comment on the line order in the tests applies as in the other PR, and I left another comment on the description, but for both of those, we can decide not to make changes there and merge this as is IMO.

mlir/include/mlir/Dialect/DXSA/IR/DXSADoubleTypeConversionOps.td
tagolog: I've created an issue to address this later https://github.com/access-softek/llvm-project/issues/204

PR #188, opened 2026-06-20
[mlir][dxsa] Migrate dcl operands to the DstOperandAttr

tagolog:
Migrates `dcl_*` operations off the legacy operand representations (the SSA-value `dxsa.operand` form and the `#dxsa.inline_operand` attribute) onto the unified `#dxsa.dst_operand` attribute, then deletes the now-unused inline operand attribute and type.
After this change all `dcl*` operands share one textual form (`v<0, <x>>`, `o<0>`, `g<0>`, ...), matching the arithmetic ops.


mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -1705,6 +1634,19 @@ class Parser {
                                        fields->values, fields->values64);
   }
 
+  // Get plain plain immediates with no relative indices.
```
asavonic: Typo

asavonic (approved):

mlir/lib/Target/DXSA/BinaryParser.cpp

PR #190, opened 2026-06-21
[mlir][dxsa] Add dadd, ddiv, dmax, dmin, dmul and drcp instructions

tagolog:
Example:
  dxsa.dadd r<0>, r<1>, r<2>
  dxsa.ddiv r<0>, r<1>, r<2>
  dxsa.dmax r<0>, r<1>, r<2>
  dxsa.dmin r<0>, r<1>, r<2>
  dxsa.dmul r<0>, r<1>, r<2>
  dxsa.drcp r<0>, r<1>


mlir/include/mlir/Dialect/DXSA/IR/DXSADoubleArithOps.td
```diff
@@ -0,0 +1,308 @@
...
+    dxsa.dadd r<0, <x, y>>, r<1, <z, w, x, y>>, -r<2>
+    ```
+  }];
+}
+
+//===----------------------------------------------------------------------===//
+// dxsa.dadd_sat
+//===----------------------------------------------------------------------===//
+
+def DXSA_DaddSat : DXSA_BinaryOp<"dadd_sat"> {
```
hvdijk: I'm seeing in the documentation that `_sat` is an instruction modifier, these aren't actually separate instructions? In that case, would it make sense to find a way to avoid duplicating almost every instruction?

mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -2336,18 +2372,29 @@ class Parser {
             beginOffset, instructionLengthInTokens)))
       return failure();
 
-    Instruction dclInstruction;
-    auto parseResult =
-        parseDclInstruction(*opcodeToken0, getLocation(), dclInstruction);
-    if (parseResult.has_value()) {
+    Instruction instruction;
+    using TryParseFunction = std::function<OptionalParseResult(Instruction &)>;
+    const TryParseFunction tryParseFunctions[] = {
```
asl: don't you want to do differently?

tagolog: Gone with reabse.

mlir/include/mlir/Dialect/DXSA/IR/DXSADoubleArithOps.td
tagolog: We decided to support `_sat` as a separate instruction (s) and a lot of instructions already implemented in this manner.

hvdijk: That seems inconsistent with the logic @asl gave in #194 against structured MLIR: "This is a bytecode dialect." In the bytecode, these aren't separate opcodes, so if we're modelling the MLIR after what the bytecode looks like, they shouldn't be separate instructions.

If it's decided that we're doing things this way then I'm of course not going to hold up this PR over it; I'm just leaving this comment to make it clear to anyone reading that it's intentional.

tagolog: @asl what do you mean?

tagolog: The final decision is to go with _sat as separate instructions

asavonic: For the record, I think the motivation is to keep instructions consistent with what we already added. I also have concerns over multiple optional attributes, but in this case, `sat` attribute may not be optional and have two values instead: `sat` and `no_sat`.

asavonic (approved):
LGTM.

PR #191, opened 2026-06-21
[mlir][dxsa] Add integer arithmetic instructions

tagolog:
Example:
 dxsa.iadd r<0>, -r<1>, -r<2>
 dxsa.imax r<0>, r<1>, r<2>
 dxsa.imin r<0>, r<1>, r<2>
 dxsa.ineg r<0>, r<1>
 dxsa.umax r<0>, r<1>, r<2>
 dxsa.umin r<0>, r<1>, r<2>

### DIFF
--- a/mlir/lib/Target/DXSA/BinaryParser.cpp
+++ b/mlir/lib/Target/DXSA/BinaryParser.cpp
@@ -2392,6 +2392,19 @@ public:
       return PLAIN_OP(Uge, 1, 2, HasPreciseAttr::Yes);
     case D3D10_SB_OPCODE_ULT:
       return PLAIN_OP(Ult, 1, 2, HasPreciseAttr::Yes);
+    // Integer arithmetic instructions
+    case D3D10_SB_OPCODE_IADD:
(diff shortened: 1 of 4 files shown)


mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -2357,16 +2370,28 @@ class Parser {
...
+          return parseDclInstruction(*opcodeToken0, getLocation(), instruction);
+        },
+        [&](Instruction &instruction) {
+          return parseIntArithInstruction(
+              opcode, beginOffset, instructionLengthInTokens,
+              modifier.preciseMask, getLocation(), instruction);
+        },
+    };
+    for (const auto &tryParse : tryParseFunctions) {
+      auto parseResult = tryParse(instruction);
```
hvdijk: The same comment as in #194 applies here, except I think this PR is actually the one it came from so it's probably better to address that here?

tagolog: Gone with rebase.

hvdijk (approved):

PR #192, opened 2026-06-21
[mlir][dxsa] Add imm_atomic_alloc and imm_atomic_consume instructions

tagolog:
Example:
  dxsa.imm_atomic_alloc r<0, <x>>, u<0>
  dxsa.imm_atomic_consume r<0, <x>>, u<0>

### DIFF
--- a/mlir/test/Target/DXSA/atomic_ops.test
+++ b/mlir/test/Target/DXSA/atomic_ops.test
@@ -187,3 +187,17 @@
 // CHECK-NEXT:   dxsa.imm_atomic_exch r<0, <x>>, u<0>, r<1>, r<2>
 // CHECK-NEXT: }
 0x090000b8, 0x00100012, 0x00000000, 0x0011e000, 0x00000000, 0x00100e46, 0x00000001, 0x00100e46, 0x00000002
+
+// -----
+
+// CHECK-LABEL: dxsa.module {
+// CHECK-NEXT:   dxsa.imm_atomic_alloc r<0, <x>>, u<0>
+// CHECK-NEXT: }
+0x050000b2, 0x00100012, 0x00000000, 0x0011e000, 0x00000000
+
+// -----
+
+// CHECK-LABEL: dxsa.module {
+// CHECK-NEXT:   dxsa.imm_atomic_consume r<0, <x>>, u<0>
+// CHECK-NEXT: }
+0x050000b3, 0x00100012, 0x00000000, 0x0011e000, 0x00000000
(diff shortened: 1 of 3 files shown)


mlir/test/Target/DXSA/atomic_ops.test
```diff
@@ -0,0 +1,14 @@
+// RUN: mlir-translate --split-input-file --import-dxsa-hex %s | FileCheck %s
+// RUN: mlir-translate --split-input-file --import-dxsa-hex %s | mlir-opt --split-input-file --verify-roundtrip
+
+// CHECK-LABEL: dxsa.module {
+// CHECK-NEXT:   dxsa.imm_atomic_alloc r<0, <x>>, u<0>
+// CHECK-NEXT: }
+0x050000b2, 0x00100012, 0x00000000, 0x0011e000, 0x00000000
```
hvdijk, line 196: For consistency with other tests these could be reversed, but I don't mind if you'd rather leave this as it is now.
```suggestion
0x050000b2, 0x00100012, 0x00000000, 0x0011e000, 0x00000000
// CHECK-NEXT: }
```

mlir/test/Target/DXSA/atomic_ops.test
```diff
@@ -0,0 +1,14 @@
...
+
+// -----
+
+// CHECK-LABEL: dxsa.module {
+// CHECK-NEXT:   dxsa.imm_atomic_consume r<0, <x>>, u<0>
+// CHECK-NEXT: }
+0x050000b3, 0x00100012, 0x00000000, 0x0011e000, 0x00000000
```
hvdijk, line 203:
```suggestion
0x050000b3, 0x00100012, 0x00000000, 0x0011e000, 0x00000000
// CHECK-NEXT: }
```

hvdijk (approved):

mlir/test/Target/DXSA/atomic_ops.test
tagolog: Good point on consistency. I'd prefer keeping it, though.

In the previous multi-instruction tests (with NO `--split-input-file` option) all instructions share a single module, so the hex is one continuous token stream. There it reads naturally to glue each token line right under its CHECK-NEXT op line, interleaving the expected output with the input as the stream progresses.

Here, with `--split-input-file` option, every instruction is its own SHEX chunk, i.e. its own module. Keeping the whole expected module together as one CHECK block `dxsa.module { ... op ... }` and pulling the hex out into a single contiguous block below draws a clean line between "expected IR" and "input tokens".

hvdijk: We also have that in other single-instruction tests, but I think I agree with you and if we consistently use the style you're using here, that sounds good to me.

PR #198, opened 2026-06-24
[mlir][dxsa] Add topology instructions

dzhidzhoev:
Example:
dxsa.cut
dxsa.emit
dxsa.emit_then_cut
dxsa.emit_stream 1
dxsa.cut_stream 3
dxsa.emit_then_cut_stream 2

### DIFF
--- a/mlir/include/mlir/Dialect/DXSA/IR/DXSAOpBase.td
+++ b/mlir/include/mlir/Dialect/DXSA/IR/DXSAOpBase.td
@@ -28,6 +28,12 @@ class DXSA_Op<string mnemonic, list<Trait> traits = []> :
 // DXSA shared bases for ops with inline operands
 //===----------------------------------------------------------------------===//
 
+class DXSA_NoOperandOp<string mnemonic> : DXSA_Op<mnemonic> {
+  let arguments = (ins);
+  let results = (outs);
+  let assemblyFormat = "attr-dict";
+}
+
 class DXSA_UnaryOp<string mnemonic> : DXSA_Op<mnemonic> {
   let arguments = (ins
       DXSA_DstOperandAttr:$dst,
@@ -72,6 +78,12 @@ class DXSA_MultiplyAddOp<string mnemonic> : DXSA_Op<mnemonic> {
   let results = (outs);
   let assemblyFormat =
       "(`precise` $precise^)? $dst `,` $lhs `,` $rhs `,` $acc attr-dict";
+}
+
+class DXSA_GsStreamIndexOp<string mnemonic> : DXSA_Op<mnemonic> {
+  let arguments = (ins ConfinedAttr<I32Attr,
+      [IntNonNegative, IntMaxValue<3>]>:$index);
+  let assemblyFormat = [{ $index attr-dict }];
 }
 
 #endif // MLIR_DIALECT_DXSA_IR_DXSAOPBASE
--- a/mlir/include/mlir/Dialect/DXSA/IR/DXSATopologyOps.td
+++ b/mlir/include/mlir/Dialect/DXSA/IR/DXSATopologyOps.td
@@ -1,0 +1,154 @@
+//===- DXSATopologyOps.td - DXSA topology ops --------------*- tablegen -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// Topology instructions of the DXSA dialect.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef MLIR_DIALECT_DXSA_IR_DXSATOPOLOGYOPS
+#define MLIR_DIALECT_DXSA_IR_DXSATOPOLOGYOPS
+
+include "mlir/Dialect/DXSA/IR/DXSAOpBase.td"
+
+//===----------------------------------------------------------------------===//
+// dxsa.emit
+//===----------------------------------------------------------------------===//
+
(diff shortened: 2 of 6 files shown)


mlir/include/mlir/Dialect/DXSA/IR/DXSATopologyOps.td
```diff
@@ -0,0 +1,154 @@
...
+    dxsa.cut
+    ```
+  }];
+}
+
+//===----------------------------------------------------------------------===//
+// dxsa.emit_then_cut
+//===----------------------------------------------------------------------===//
+
+def DXSA_EmitThenCut : DXSA_NullaryOp<"emit_then_cut"> {
```
hvdijk: Microsoft call this [`emitThenCut`](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/emitthencut--sm4---asm-) rather than `emit_then_cut`, likewise for [`emitThenCut_stream`](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/emitthencut-stream--sm5---asm-), is it possible to use their name or does it need to be this?

tagolog: Instruction names in Microsoft's specification are formed according to different naming rules, rather than following a single uniform convention. We decided to use snake_case for operation names to maintain consistency across the dialect.

mlir/include/mlir/Dialect/DXSA/IR/DXSATopologyOps.td
```diff
@@ -0,0 +1,154 @@
+//===- DXSTopologyOps.td - DXSA topology ops ---------------*- tablegen -*-===//
```
tagolog:
```suggestion
//===- DXSATopologyOps.td - DXSA topology ops ---------------*- tablegen -*-===//
```

mlir/test/Target/DXSA/topology.test
```diff
@@ -0,0 +1,51 @@
...
+// CHECK-NEXT:   dxsa.emit
+0x01000013
+
+// CHECK-NEXT:   dxsa.emit_then_cut
+0x01000014
+
+// CHECK-NEXT:   dxsa.emit_stream 0
+0x03000075, 0x00110000, 0x00000000
+
+// CHECK-NEXT:   dxsa.emit_stream 1
```
tagolog: Redundant: all three stream ops decode the index through the same path. Maybe we can keep happy path for boundaries only (0 and 3)?

mlir/test/Target/DXSA/topology.test
```diff
@@ -0,0 +1,51 @@
...
+0x03000075, 0x00110000, 0x00000001
+
+// CHECK-NEXT:   dxsa.emit_stream 2
```
tagolog: See above.

mlir/test/Target/DXSA/topology.test
```diff
@@ -0,0 +1,51 @@
...
+0x03000075, 0x00110000, 0x00000002
+
+// CHECK-NEXT:   dxsa.emit_stream 3
+0x03000075, 0x00110000, 0x00000003
+
+// CHECK-NEXT:   dxsa.cut_stream 0
+0x03000076, 0x00110000, 0x00000000
+
+// CHECK-NEXT:   dxsa.cut_stream 1
```
tagolog: See above.

mlir/test/Target/DXSA/topology.test
```diff
@@ -0,0 +1,51 @@
...
+0x03000076, 0x00110000, 0x00000001
+
+// CHECK-NEXT:   dxsa.cut_stream 2
```
tagolog: See above.

mlir/test/Target/DXSA/topology.test
```diff
@@ -0,0 +1,51 @@
...
+0x03000076, 0x00110000, 0x00000002
+
+// CHECK-NEXT:   dxsa.cut_stream 3
+0x03000076, 0x00110000, 0x00000003
+
+// CHECK-NEXT:   dxsa.emit_then_cut_stream 0
+0x03000077, 0x00110000, 0x00000000
+
+// CHECK-NEXT:   dxsa.emit_then_cut_stream 1
```
tagolog: See above.

mlir/test/Target/DXSA/topology.test
```diff
@@ -0,0 +1,51 @@
...
+0x03000077, 0x00110000, 0x00000001
+
+// CHECK-NEXT:   dxsa.emit_then_cut_stream 2
```
tagolog: See above.

tagolog (changes requested):

mlir/test/Target/DXSA/topology.test
dzhidzhoev: Sure, fixed that.

mlir/include/mlir/Dialect/DXSA/IR/DXSAOpBase.td
```diff
@@ -28,6 +28,12 @@ class DXSA_Op<string mnemonic, list<Trait> traits = []> :
 // DXSA shared bases for ops with inline operands
 //===----------------------------------------------------------------------===//
 
+class DXSA_NullaryOp<string mnemonic> : DXSA_Op<mnemonic> {
```
hvdijk: This is identical to what is in #194; I'm not sure which PR had it first and where the review comment should go. Regardless, the same comment applies here.

hvdijk (changes requested):

mlir/include/mlir/Dialect/DXSA/IR/DXSAOpBase.td
tagolog: We decided to proceed with the simple solution and not merge the complicated version with customizable PlainOp from #201

tagolog (approved):

PR #206, opened 2026-06-27
[mlir][dxsa] Refine tests for condition instructions

tagolog:
Merge per-op condition/{eq,ge,lt,ne,ieq,ige,ilt,ine,uge,ult}.test files into a single condition_ops.test, and reduce each instruction to two cases: a composite mask+swizzle (plus neg/neg_abs for float ops) and a precise variant.


asavonic (approved):
LGTM.

PR #207, opened 2026-06-27
[mlir][dxsa] Add mov, dmov, movc and swapc instructions

tagolog:
Example:
 dxsa.mov r<0>, r<1>
 dxsa.dmov r<0, <x, y>>, -|r<1, <z, w, x, y>>|
 dxsa.dmovc r<0>, r<1>, r<2>, r<3>
 dxsa.swapc r<0>, r<1>, r<2>, r<3>, r<4>


mlir/include/mlir/Dialect/DXSA/IR/DXSAMoveOps.td
```diff
@@ -0,0 +1,207 @@
...
+    dxsa.mov r<0, <x, y>>, r<1, <y, z, x, w>>
+    ```
+  }];
+}
+
+//===----------------------------------------------------------------------===//
+// dxsa.mov_sat
+//===----------------------------------------------------------------------===//
+
+def DXSA_MovSat : DXSA_UnaryOp<"mov_sat"> {
```
hvdijk, line 43: The question of how we want to deal with `_sat` (from #190) is unanswered, and would affect this PR too.

mlir/include/mlir/Dialect/DXSA/IR/DXSAMoveOps.td
```diff
@@ -0,0 +1,207 @@
...
+    The `dxsa.swapc` operation performs a component-wise conditional swap of
+    the values in `$src1` and `$src2`:
+
+    ```
+    $dst0 = ($condition != 0) ? $src2 : $src1
+    $dst1 = ($condition != 0) ? $src1 : $src2
+    ```
+
+    Both destinations are updated atomically with respect to the source
+    values, so reading and writing the same register is well-defined.
```
hvdijk, line 183: It's well-defined but with different semantics than the above pseudocode and this description suggest. This description suggests `$dst0` is updated atomically, and then `$dst1` is updated atomically. That's not what happens and not how Microsoft document it (they introduce a `temp` variable to make the meaning clear).

asavonic (approved):
LGTM. Please address Harald's comment on the doc.
Regarding `_sat`, it is the same issue as in #190. Let's keep instructions consistent for now, and if we implement `sat` as an attribute, we can refactor this patch.

mlir/include/mlir/Dialect/DXSA/IR/DXSAMoveOps.td
tagolog: `_sat` is a part of the operation name for now.

PR #208, opened 2026-06-27
[mlir][dxsa] Add dfma instruction

dzhidzhoev:
Example:
dxsa.dfma <r0>, r<0>, r<1>, r<2>
dxsa.dfma_sat <r0>, r<0>, r<1>, r<2>

Stacked on https://github.com/access-softek/llvm-project/pull/190.

### DIFF
--- a/mlir/include/mlir/Dialect/DXSA/IR/DXSADoubleArithOps.td
+++ b/mlir/include/mlir/Dialect/DXSA/IR/DXSADoubleArithOps.td
@@ -305,4 +305,52 @@ def DXSA_DRcpSat : DXSA_UnaryOp<"drcp_sat"> {
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// dxsa.dfma
+//===----------------------------------------------------------------------===//
+
+def DXSA_DFma : DXSA_MultiplyAddOp<"dfma"> {
+  let summary = "component-wise double-precision multiply-add";
+  let description = [{
+    The `dxsa.dfma` operation computes the component-wise double-precision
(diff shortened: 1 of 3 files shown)


mlir/include/mlir/Dialect/DXSA/IR/DXSADoubleArithOps.td
```diff
@@ -13,7 +13,7 @@
 #ifndef MLIR_DIALECT_DXSA_IR_DXSADOUBLEARITHOPS
 #define MLIR_DIALECT_DXSA_IR_DXSADOUBLEARITHOPS
 
-include "mlir/Dialect/DXSA/IR/DXSAOpBase.td"
+include "mlir/Dialect/DXSA/IR/DXSAFPArithOps.td"
```
tagolog: Maybe we just move `DXSA_MultiplyAddOp` to the `DXSAOpBase.td`? This seems a better solution rather then including FP ops to double.

tagolog (changes requested):

mlir/include/mlir/Dialect/DXSA/IR/DXSADoubleArithOps.td
dzhidzhoev: Thank you, that makes sense. I did that in https://github.com/access-softek/llvm-project/pull/210. I will update it here after https://github.com/access-softek/llvm-project/pull/210 and https://github.com/access-softek/llvm-project/pull/190 get merged.

tagolog: Fixed with rebase

tagolog (approved):

PR #212, opened 2026-06-29
[mlir][dxsa] Add sync instruction

tagolog:
Example:
 dxsa.sync <uav_global|threads>

### DIFF
--- a/mlir/lib/Target/DXSA/BinaryParser.cpp
+++ b/mlir/lib/Target/DXSA/BinaryParser.cpp
@@ -566,6 +566,11 @@ public:
     return dxsa::DclGlobalFlags::create(builder, loc, flagsAttr);
   }
 
+  Instruction buildSync(dxsa::SyncFlags flags, Location loc) {
+    auto flagsAttr = dxsa::SyncFlagsAttr::get(builder.getContext(), flags);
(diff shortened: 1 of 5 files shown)


mlir/lib/Target/DXSA/BinaryParser.cpp
```diff
@@ -2477,6 +2498,14 @@ class Parser {
       return PLAIN_OP(ImmAtomicAlloc, 1, 1, HasPreciseAttr::No);
     case D3D11_SB_OPCODE_IMM_ATOMIC_CONSUME:
       return PLAIN_OP(ImmAtomicConsume, 1, 1, HasPreciseAttr::No);
+    case D3D11_SB_OPCODE_SYNC: {
+      auto sync = parseSync(*opcodeToken0, getLocation());
+      FAILURE_IF_FAILED(sync);
+      if (failed(
+              verifyInstructionLength(beginOffset, instructionLengthInTokens)))
```
asavonic: Can we move the length check to `parseSync`,  so it is more in line with other operations?

asavonic (approved):
LGTM.

mlir/lib/Target/DXSA/BinaryParser.cpp

